PR #1, opened 2024-03-25
feat: clean-up package structure and partial TypeScript fixes

grabbou:
This PR is rather long. I am not done yet, but I figured it was a good moment to wrap it up as I am half way through it. There will be some follow-ups, especially around fixing ESLint / TypeScript that I will tackle later on.

This PR includes the following:

1. Removed some stuff that we are not using right now:  `turborepo`, `manypkg`, `node-gyp` and more
2. Regrouped dependencies to match what current package is using
3. Cleaned up extraneous workspace dependencies: should not contain anything React Native / Expo related.
4. Removed resolutions for React versions. I don’t think we need it. We can always check with `yarn-dedupe` or inspect the lock file. If we end up with multiple versions of React, there must be a reason, so it’s better to debug that reason first before jumping to fixing. We already have an auxiliary script `check-deps` to ensure correct versioning within the monorepo.
5. Run `check-deps` to fix some issues. Otherwise, fresh project didn’t want to build.
6. Moved some gitignore files to the top-level and removed inner ones.
7. Removed `.nvmrc` (no clear guidelines we use it actually. I e.g. use something else) and moved Node engine requirement to the top-level package.
8. I haven’t tested release-related code yet. We will figure this out later, but I am leaving placeholder for now at the top-level only.
9. Changed `yarn build` to not run on every install. It is not necessary. We’re using source files anyways during development.
10. Got rid of tamlgui-build in favour of unified build pipeline with react-native-bob for simplicity
11. Renamed `app` to `@react-native-templates/app` (for now) and re-grouped its structure to match other packages (it's pending refactoring anyway)
12. Removed babel plugin to rewrite imports inside `rise-mobile` and using normal module system at this point. Had to re-export some members (with * for now) from the app package


grabbou:
We can get back to using `yarn` once I resolve some build issues, but I think for time being, defaulting to Node LTS and bundled npm is not a bad idea at all.

grabbou:
Additional changes:
- Removed `react-native-builder-bob` and will be running a single `tsc` script to build all packages
- Cleaned up duplicated Tamagui configuration and invalid Next.js configuration (we now import web extension over normal one, so React Native code does not get loaded)
- Added `CONTRIBUTING.md` file with a starting point to document ways of working on this project and helpful commands
- Some other smaller changes that I lost track of

PR #2, opened 2024-03-29
feat: refactor core package and provide basic unit tests

grabbou:
The purpose of this PR is to improve the core package by adding unit tests and providing full TypeScript coverage. While doing so, certain optimisation opportunities were identified, which resulted in a slightly reorganised structure.

Most importantly:
- function renderElement and renderProp was merged into single function because prop can also contain an element (e.g. "header={<Header />}" in typical React code. Provided test case for this scenario
- extracted "refs" logic out to a separate, higher-order component. You can use Template to work with refs, or BaseTemplate to just render pure data w/o refs. Allows for better unit testing ,but also provides a light-weight alternative.
- (temporarily) removed if conditionals checking for data integrity as we have to decide whether to use `zod` or bring back a very minimal JavaScript code to verify types are correct during runtime
- for error handling, switched to throwing errors and using custom error boundary to provide placeholder for an errored component. This is TBD.
- removed dependency on React Native, this package is React only from now on and can be used cross-platform. "app" package will import "core" and inject React Native specific views (e.g. from Tamagui). This will happen in a separate PR.
- renamed and unified some naming patterns (data state, data values, data store, data source)
- added snapshot tests for some of the supported use cases
- added more tests for refs


packages/core/src/__tests__/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,59 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`should accept component as a prop 1`] = `
+<DocumentFragment>
+  <section>
+    <header>
+      <div
+        data-testid="$heading.$props.$header"
```
grabbou, line 8: Couldn't find a better way to verify keys work properly and since this is crucial part of the library (I have some questions around it btw), I figured it might be better if we explicitly check for it.

Overall, not sure we're going to use testing library going forward, depends whether it's possible to check props/keys/whatever from user defined components as well

ericvicenti:
Merged!

PR #3, opened 2024-04-11
feat: add tamagui and rise ui packages

grabbou:
Moving from core to second package, which is the ui. The idea is to provide a good base collection of UI components that can be used by developers out of the box and Tamagui seems like a good viable option with cross-platform support and quite comprehensive components collection.

We may consider using RNW or RNSD at some point, especially after latter stabilises. For now, let's default to something that we're going to use within the showcase app itself.

Changelog:
- Added `ui-tamagui` package with basic components as per above message
- Added `ui-rise` (custom higher order components). Some of them were renamed (with a prefix), will need to update backend code in a follow-up commit
- Broke down `demoComponents.tsx` file so that each component resides in a separate file. 
- Removed "prop || default" in favour of using Zod's defaults. `Validator` can tweak props, so it's safe.
- Removed `Button` component and added `wrapEvents` function to quickly proxy functions such as `onPress` to `onTemplateEvent`.
- Removed `demo-ui` - prefer using tamagui directly. `Toast` was moved to `app/provider` since it's custom code that is application specific
- Import tamagui configuration from "app" package (where you get components from anyway) instead of re-exporting from `demo-ui`
- Removed shorthands (`f` or `ai`) so it's easier to understand Tamagui props for other contributors. I am also preparing to remove built-in tamagui configuration and replace with a smaller/simpler one that may not include prefixes out of the box.

To do:
- unit tests
- run the app and make sure it works
- there are some todos in the code but probably better to handle as a sep. PR
- refactor/clean-up components in `ui-rise` itself. creating component libraries is tricky and this package is used internally right now. However, it might be easier for end-users to use this as it greatly simplifies some of the Tamagui components configurations. We could perhaps rename it to `ui-tamagui-extended`.


ericvicenti:
I'm excited for these new UI packages! The iOS app is broken so lets not merge into main yet

grabbou:
Good. Let me rebase against `next` and merge so we can move forward.

PR #5, opened 2024-04-12
feat: new event handling

grabbou:
There are a few key changes in this PR regarding event handling logic:
- removal of previously introduced `wrapEvents` utility function. Now, presence of our library is completely transparent and there's no modifications that have to be done to render any component with any event handlers
- to configure an event handler, simply do as follows:
  ```tsx
  { 
    $: 'component', 
    component: 'TouchableOpacity', 
    props: { onPress: { $: 'event' } }
  }
  ```
  in this bare scenario, our library will convert onPress "object" to a function and pass that function to component instead. When Touchable is pressed, there will be an event originating from that component. This is bare format. More options are available (as explained below).
- you can see this working quite well in `ui-rise` where we have custom higher-order components. Even though event handler is defined as an object with Zod, TypeScript does not complain because "TemplateComponentProps" translates them all to a function
- event handler can have an `action` in case you want to trigger something that is supported by the environment (e.g. Rise) or your custom code. This is utility property for your use case. It allows you to identify an event.
- the shape of event is as follows: 
  ```tsx
  export type TemplateEvent = {
    target: {
      key?: string
      path: string
    }
    name: string
    action?: string
    payload: any[]
  }
  ```
  You have a few ways to listen to an event. You can either check the "action", which is proprietary name of your event. You can also check the target: key - if defined, this is the key of the element as set by you in your JSON, or path - which is the location in the render tree where the event originated from. That way if multiple buttons are rendered, you can locate the right one. Note that `action` is optional, it's up to you whether you want to use `name` & `target` or `action` or both.
  Payload will be arguments as originally received by the handler. Name is the name of the prop that triggered the event, e.g. "onPress" or "onDragEnd"
- In production `TemplateEvent` might look like this:
  ```tsx
    const event = { target: { 'myHeaderButton', path: '<path_to_component_in-tree>' }, name: 'onPress', action: 'navigation:goBack', payload: [] }
  ``` 
  for an event that was defined as follows:
  ```tsx
  { $: 'component', type: 'TouchableOpacity', props: { onPress: { $: 'event', action: 'navigation:goBack' }}}
  ```
- You can use `EventDataStateSchema` (Zod) to parse arguments inside `validate` to make sure your component receives correct props. In the future, I will add the same for ComponentDataState, so that we can also validate that correct shape was passed (e.g. if your component accepts components as props).
- Thanks to that, I was able to clean-up some annoying workarounds, such as having to avoid passing `onTemplateEvent` to "div" to avoid warnings: 
Screenshot 2024-04-12 at 17 57 33.
Now, we can just pass through div props.
- Separated JSONValue from DataState so that our custom objects (with "$" key) are not getting confused with standard JSONValues that can be passed from the API


ericvicenti (approved):
Looks good, `$: 'event'` seems like a big improvement!

Only downside to this approach, is that currently the Uprising app expects events to automatically be created even when they are not defined from the template definition

So we would previously expect a button to always emit an onPress event, but now the event will only be created when `props: { onPress: { $: 'event' }}` is set.

But this probably isn't that big of a deal, just more verbose code on the server for now until we build nicer abstractions

grabbou:
> So we would previously expect a button to always emit an onPress event, but now the event will only be created when props: { onPress: { $: 'event' }} is set.

Yeah, this is valid point. Initially, I ended up with similar implementation, but quickly realised that it won't be scalable to listen to all events, especially that certain components can alter their behavior when listeners are set (e.g. ScrollView). I think for now a bit more verbosity will do, we can always revisit with a higher order abstraction that does that automatically.

For example, `button()` function (similar to section).

PR #8, opened 2024-04-14
feat: update web socket server to new event handling 

grabbou:
Note: the diff is large b/c of `lock file` updates. 

- Updates web socket server to new event handling
- Moved web socket code into reusable package
- Moved onEvent from Template to inside, as it's less error-prone (use "interceptEvent" instead).
- Removed ESBuild, ESBuild Runner and using TSX instead to run TypeScript in dev mode. Since TSX has watch mode, Nodemon is not necessary. To build for production, we will just run `tsc` (TypeScript compiler, as other code). This will be handled later, when we do PR for production deployments
- Adds TypeScript types and updates signatures where necessary
- We now also look for $: event in nested objects when given as props, e.g:
  ```ts
   { $: 'component', type: 'List', props: { items: [{ onPress: { $: 'event' }}]}}
   ```
   is now going to be possible and you will be able to call `props.items[0].onPress()` to trigger an event.
- Simplified some code (instead of has() && get(), just getting from a map)
- Changed event payload from array to single argument. 99% of the cases, event handlers only receive one argument, so doesn't make sense to do (...args: []). Unnecessary boilerplate to write event.payload[0] all the time. For that 1% cases where multiple arguments are passed, one can always create a higher order component and handle that case in the user space.
- Renamed components to use prefix "Rise" as per "ui-rise" (updated in the other commit)
- Not serialising event payload and not sending to the server over WebSocket connection IF it is a NativeEvent. These are quite heavy cyclic structures and provide little to no value on the server. 
- Added better error message when validation fails
- Added type of component that triggered the event, so we can check for specific component events (e.g. `event.target.component === 'RiseSliderField'` will be true for all events coming from that type of a component)
- Replaced event handler: `action === 'reverse' && name === 'onValueChange'` (name of action and type of event) to just name of action `action === reverse `. This is probably left over from code where it used to check for `key === 'reverse'` (which makes sense - specific event triggered by a specific element). You can see in the code switches have same key as the action: 
   ```tsx
   {
      $: 'component',
      key: 'reverse',
      component: 'SwitchField',
      props: {
        value: state?.params?.reverse || false,
        label: 'Reverse Playback',
        onValue: {
          $: 'event',
          action: ['updateMedia', mediaPath, 'reverse'],
        },
      },
    },
    ```
    Since we only check explicit action that is fired on explicit event, there's no need to check for the name of the event or its key. In fact, we could remove the "key" as its optional and not really necessary in this case.

Unit tests will be added in a separate PR.


grabbou:
To run, recommended to remove `node_modules` and do `npm install` once again. Clean up everything. Rebuild the mobile project.

packages/app/src/fields.tsx
```diff
@@ -35,43 +34,48 @@ const FieldError = ({ message }: { message?: string }) => {
...
   const themeName = useThemeName()
-  const id = useId()
   const disabled = isSubmitting
-  const isPrettyName = !isURL && !isEmail && uniqueId !== 'path'
+
   return (
     <Theme name={error ? 'red' : themeName} forceClassName>
       <Fieldset>
         {!!label && (
-          <Label theme="alt1" size={props.size || '$3'} htmlFor={id}>
```
grabbou, line 54: I don't think we need this "htmlFor" at this point and `useId()` does not work for some reason very well within ts-form. It produces duplicate IDs. From my research, this might be related to useId() when used inside a fragment.

apps/next/tamagui.config.ts
```diff
@@ -1,4 +1,7 @@
-import { tamaguiConfig } from '@react-native-templates/app'
```
grabbou, line 1: to fix the eg-preview and also make it independent of React native Templates, I removed import of Tamagui configuration since it is not really relevant. 

That way, we don't have to set up Expo, transpile React Native or alias packages, as there won't be any React Native dependencies pulled into the build by accident.

grabbou:
Working proof: 

Screenshot 2024-04-17 at 11 21 42

packages/ws-client/src/data-source.ts
```diff
@@ -0,0 +1,127 @@
...
+      return
+    }
+    send({
+      $: 'sub',
+      keys,
+    })
+  })
+
+  rws.onmessage = (eventData) => {
+    const event = JSON.parse(eventData.data) as ServerWebsocketMessage
```
ericvicenti, line 60: we should validate that the message is correct, using zod or something. this is technically not safe

packages/core/src/__tests__/refs.test.tsx
```diff
@@ -190,15 +181,11 @@ it('should manage subscription to stores referenced by refs', () => {
...
       }
     },
+    sendEvent: jest.fn(),
   }
 
   const element = render(
-    <Template
-      components={BUILT_IN_COMPONENTS}
-      dataSource={dataSource}
-      onEvent={jest.fn()}
```
ericvicenti, line 199: hm I will miss having onEvent directly on the template component. It feels nice that this can be separated from the data source.

packages/app/src/fields.tsx
ericvicenti: oh hmm. Well the `htmlFor` is needed so the correct input gets focused when you tap on the label.

apps/next/tamagui.config.ts
ericvicenti: ok seems reasonable

apps/server/package.json
```diff
@@ -1,16 +1,30 @@
...
+  "bugs": {
+    "url": "https://github.com/ericvicenti/react-native-templates/issues"
+  },
+  "homepage": "https://github.com/ericvicenti/react-native-templates#readme",
   "scripts": {
-    "start": "esr server.ts",
-    "start:demo": "nodemon -e js,ts,tsx --exec \"esr demo-server.ts\"",
-    "start:dev": "nodemon -e js,ts,tsx",
-    "convert-videos": "esr eg-video-conversion.ts"
+    "start": "tsx server.ts",
```
ericvicenti, line 19: this will be slower than `esbuild runner` but no big deal

packages/app/src/connection-form.tsx
```diff
@@ -15,7 +15,7 @@ export function ConnectionForm({
   submitButton,
 }: {
   onSubmit: (values: z.infer<typeof connectionFormSchema>) => void
-  defaultValues: z.infer<typeof connectionFormSchema>
```
ericvicenti, line 18: I think we want to keep the z.infer here so the values are type safe. AppForm has type of `any` according to TS on my computer, so using it to extract ComponentProps defaultValues probably isn't working

ericvicenti (commented):
how is it error prone if there is onEvent prop on the `<Template>` component?

grabbou:
> how is it error prone if there is onEvent prop on the <Template> component?

In current implementation, `onEvent` can be set on a <BaseTemplate /> component but cannot be set on a <Template /> component. <Template /> component is meant to be used with a Data Store and Data Store does have `sendEvent` that we want to call. 

If someone wants to listen to the event, they should use `interceptEvent` from Data Store. Leaving that up to the user seems like an unnecessary boilerplate that, if not set correctly, can lead to the library not functioning properly.

apps/server/package.json
grabbou: Hm, I haven't run benchmarking, but it has better interior with TypeScript, so gets all the `tsconfigs` right. I was thinking that for production deployments, we should run `tsc` and run the server with pure `node` instead.

grabbou:
> oh hmm. Well the htmlFor is needed so the correct input gets focused when you tap on the label.

Yeah, but for some reason, `useId()` does not work reliably with TS Forms, so decided to remove it for now. Since we're not rendering for the web yet, I think we can leave accessibility for later.

ericvicenti:
hmm.. well I don't think you should be required to pass events into the data source. That is why it worked this way. In fact it is kind of out of scope for a data source. If it handles events, then we should call it a "model" rather than a "data source"

packages/core/src/__tests__/refs.test.tsx
grabbou: Well, my initial thought process was that: Template is meant to be used with Data Source, because Template component has a tightly coupled relation with it (subscribing to stores and listening to changes).

If you don't want that Data Source-specific functionality, you can just use BaseTemplate directly which lets you set onEvent yourself as it's not really coupled to the implementation.

ericvicenti:
> > oh hmm. Well the htmlFor is needed so the correct input gets focused when you tap on the label.
> 
> Yeah, but for some reason, `useId()` does not work reliably with TS Forms, so decided to remove it for now. Since we're not rendering for the web yet, I think we can leave accessibility for later.

It is not just for accessibility. On mobile you want to tap the label to focus the related input. If `useId` is broken then the id should be a required prop

packages/ws-client/src/data-source.ts
grabbou: Yeah, we need to add runtime type checking to many places, including this. I will handle that in a follow-up PR.

packages/app/src/connection-form.tsx
grabbou: Ideally, this should be of whatever AppForm type is. Unfortunately, it's `any`. `z.infer<typeof connectionFormSchema>` was a bit misleading, because if `connectionFormSchema` was incompatible with the AppForm, it would still satisfy TypeScript type checking if you passed an instance of it. 

Either way, it's not really type safe, but I don't find it a big issue given it's just a single form at this point. I am happy to make adjustments if you feel strong about that one tho!

grabbou:
> It is not just for accessibility. On mobile you want to tap the label to focus the related input. If useId is broken then the id should be a required prop

Okay, I will bring that back this way or another in a follow-up commit to this PR and let you know.

packages/app/src/screens/connection.tsx
```diff
@@ -38,20 +38,26 @@ function ActiveConnectionScreen({ connection }: { connection: Connection }) {
...
+      if (action === 'navigate-back') {
+        router.back()
+        return
+      }
+      dataSource.sendEvent(event)
+    },
+    [dataSource]
+  )
+
+  return <Template components={components} dataSource={dataSource} path={path} onEvent={onEvent} />
```
ericvicenti, line 62: this new `onEvent` implementation is clean!

ericvicenti (approved):
Looking good!

PR #9, opened 2024-04-17
feat: add CI for tests and configure Jest

grabbou:
This PR configures `jest` to always run from source and configures Github Action to test it and report. I am unifying testing environment to be web. Before, as you can see from the snapshots, Tamagui was rendering React Native views (and generating warnings about <Text /> not existing in the current environment), despite testing environment being `jsdom`. This is now fixed and unified.

### DIFF
--- a/packages/ui-tamagui/src/__tests__/index.test.tsx
+++ b/packages/ui-tamagui/src/__tests__/index.test.tsx
@@ -1,7 +1,3 @@
-/**
- * @jest-environment jsdom
- */
-
 import { BaseTemplate } from '@react-native-templates/core'
 import { config } from '@tamagui/config/v3'
 import { render } from '@testing-library/react'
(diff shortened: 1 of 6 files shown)


packages/ui-tamagui/src/__tests__/index.test.tsx
```diff
@@ -40,21 +36,32 @@ it('should render a Tamagui component', () => {
...
         >
-          Welcome to Tamagui!
-        </text>
-      </view>
+          <div
+            class="_dsp-flex _ai-stretch _fd-row _fb-auto _bxs-border-box _pos-relative _mih-0px _miw-0px _fs-0"
+            data-testid="root"
+          >
+            <h1
+              class="is_H1 font_heading _col-675002279 _tt-1440318557 _ff-299667014 _fow-1366436877 _ls-905095908 _fos-1477259397 _lh-1677663454 _dsp-inline _bxs-border-box _ww-break-word _whiteSpace-normal _mt-0px _mr-0px _mb-0px _ml-0px _ussel-auto"
```
ericvicenti, line 50: I think the snapshot was easier to review when it was printing out the react native definition.

With all these obscure class names its hard to see what is happening. Also maybe it is possible the class names change for some reason but the styles still stay the same?

ericvicenti (approved):
Exciting to get CI running!!

packages/ui-tamagui/src/__tests__/index.test.tsx
grabbou: Yeah, needs further DD for sure. React Native one was a bit easier, but not really working. To make it work, we would need to switch to a Native Testing Library.

I think we can remove class="" from the snapshot to reduce boilerplate. I will look into it in a follow-up if you don't mind!

PR #10, opened 2024-04-18
feat: new readme

grabbou:
I expect there will be numerous edits to the readme, but I wanted to refresh it with recent updates.


ericvicenti (approved):
Great, we will continue iterating on ths

PR #11, opened 2024-04-18
feat: pkg rename

grabbou:


package.json
```diff
@@ -13,10 +14,10 @@
   "scripts": {
     "native": "cd apps/rise-mobile && yarn start",
     "ios": "cd apps/rise-mobile && yarn ios",
-    "serverw": "yarn workspace @react-native-templates/rise-server",
-    "expow": "yarn workspace @react-native-templates/rise-mobile",
+    "serverw": "yarn workspace eg-dashboard",
+    "expow": "yarn workspace @final-ui/rise-mobile",
     "vercel:install": "yarn set version 3.5 && yarn install",
-    "convert-videos": "cd apps/server && yarn convert-videos",
```
grabbou, line 19: My convention is to keep root package.json slim and move all relevant scripts to separate packages. What is yours? @ericvicenti 

These need to be updated. I'd rather move them to example/* and apps/* (unless they're already there).

packages/components/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@react-native-templates/ui-rise",
+  "name": "@final-ui/components",
```
grabbou, line 2: alternative name: `@final-ui/kit`

package.json
ericvicenti: Ok the tasks that you added are helpful enough so we can start to slim down the root package.json

ericvicenti (approved):
Excited about the new name, its very sleek!

PR #13, opened 2024-04-18
feat: add sequences to eg app

ericvicenti:


example/server/server.ts
```diff
@@ -538,6 +640,31 @@ function handleMediaEvent(event: TemplateEvent<ServerAction>): boolean {
     })
     return true
   }
+  if (action === 'addToSequence') {
+    console.log('hey wtf', action)
```
grabbou, line 644: 👀

example/server/server.ts
```diff
@@ -607,10 +734,73 @@ function handleMediaEvent(event: TemplateEvent<ServerAction>): boolean {
...
+        ...media,
+        sequence: (media.sequence || []).filter((item) => item.key !== itemKey),
+      }
+    })
+    return true
+  }
+  if (action === 'maxDuration') {
+    let duration = event.payload
+    if (Array.isArray(duration)) duration = duration[0] // slider gives us an array
+    if (duration === true) duration = 10
```
grabbou, line 752: Can duration be a boolean if sliders give an array?

grabbou: Also, I think we can safely just write `event.payload[0]`. At least that's what I've been doing in other places where this event is expected to come from the Slider. Part of its future signature I guess.

example/server/server.ts
```diff
@@ -607,10 +734,73 @@ function handleMediaEvent(event: TemplateEvent<ServerAction>): boolean {
...
+    console.log('maxDuration', event)
+  }
 
   return false
 }
 
+function goNext(mediaPath: string[]) {
+  rootMediaUpdate(mediaPath, (media) => {
+    if (media.type !== 'sequence') {
+      console.warn('goNext on non-sequence media', media)
```
grabbou, line 776: nit: missing `:`

grabbou (approved):
LGTM. Did a pass over the changes to better understand this feature. Left some (potentially) helpful comments.

packages/playground/src/screens/edit-connection.tsx
```diff
@@ -27,6 +27,7 @@ export function EditConnectionScreen() {
           update((conn) => ({ ...conn, ...values }))
           goHomeLink.onPress()
         }}
+        // @ts-ignore
```
grabbou, line 30: fixed this for you via https://github.com/final-ui/final/pull/14

PR #16, opened 2024-04-19
feat: runtime validation with `typia`

grabbou:
This PR replaces `zod` for Schema validation in certain packages to use `typia` instead. Typia is faster and is AOT, which should yield better performance and interop with existing Typescript types.

The biggest difference is going to be running from compiled code instead from source. You will have to start `npm run watch` or `npm run build` in order to make sure all local packages are compiled. That, however, should help to make sure that whatever we test locally while developing, is what end users of the library are going to see.


grabbou:
Going to close due to https://twitter.com/grabbou/status/1781423500097401023

PR #18, opened 2024-04-20
feat: improve performance of state updates

grabbou:
Before adding more validation to the library, would be good to make sure we're doing it efficiently right now.


grabbou:
I am going to close this for now but we will have to revisit at some point together with a bunch of other performance focused improvements.

PR #19, opened 2024-04-20
feat: form support 

grabbou:
This PR adds support for rendering custom form components by the server. Once completed, I plan to remove `ts-form` and `connection-form` entirely and switch to rendering `<Template />` component with form definition.

Notable change is making event handlers `async` on the client. That way, we can `await` on e.g. `onPress` that results in an update being sent to the server. If server responds with something, that can be handled (e.g. perform local update, display an error). That functionality will be added in the future iteration.

### DIFF
--- a/example/server/ui.ts
+++ b/example/server/ui.ts
@@ -177,6 +177,41 @@ function getVideoControls(mediaPath: string, state: VideoMedia, context: UIConte
     },
     {
       $: 'component',
+      component: 'RiseForm',
+      props: {
+        onSubmit: {
+          $: 'event',
+        },
+      },
+      children: [
+        {
+          $: 'component',
+          component: 'RiseTextField',
+          props: {
+            name: 'title',
+            label: {
+              $: 'component',
+              component: 'Label',
+              children: 'Layer title',
+              props: {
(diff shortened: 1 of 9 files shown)


example/server/ui.ts
```diff
@@ -175,6 +175,41 @@ function getVideoControls(mediaPath: string, state: VideoMedia, context: UIConte
...
+      props: {
+        onSubmit: {
+          $: 'event',
+        },
+      },
+      children: [
+        {
+          $: 'component',
+          component: 'RiseTextField',
+          props: {
```
grabbou, line 190: This form is obviously in the wrong place and with wrong values, but just wanted to give you an example. Let me know where this should be rendered and will update accordingly.

grabbou: Update from the discussion: We're going to build title editor for layers/sequences.

ericvicenti (approved):
Generally this looks great! Can we remove all of the async event stuff before merging this? Because that is not ready yet

grabbou:
Will move async to separate branch and work further on this by adding a Modal (might be a separate PR tho)

PR #20, opened 2024-04-20
feat(experimental): jsx support on the server

grabbou:
Preliminary support with custom TypeScript config. We would most likely create an appropriate preset for developers to use.

We're not going to migrate example server over to it (too much work), but I would like to implement the new demo server with this for the Apple review so we can start dogfooding. That should make it a bit easier to review/merge, even though is not perfect.


packages/react/src/template.tsx
```diff
@@ -95,7 +95,11 @@ export function BaseTemplate({
       throw new RenderError(`Invalid component: ${stateNode.component}`)
     }
 
-    let componentProps = stateNode.props || {}
+    let componentProps = Object.fromEntries(
+      Object.entries(stateNode.props || {}).map(([propKey, propValue]) => {
+        return [propKey, renderProp(propKey, propValue, stateNode, path)]
+      })
+    )
```
grabbou, line 102: We first render props, then we validate them. That way, your validation can check for `onValueChange: function` instead of `onValueChange: ${ event }`, leaking impl. detail into client code. Will cherry-pick as a separate PR as this is good improvement.

tsconfig.json
```diff
@@ -4,10 +4,12 @@
     "paths": {
       "@final-ui/app": ["./packages/playground/src"],
       "@final-ui/react": ["./packages/react/src"],
+      "@final-ui/react/jsx-runtime": ["./packages/react/src/jsx"],
       "@final-ui/ws-client": ["./packages/ws-client/src"],
       "@final-ui/ws-server": ["./packages/ws-server/src"],
       "@final-ui/tamagui": ["./packages/tamagui/src"],
-      "@final-ui/kit": ["./packages/kit/src"]
+      "@final-ui/kit": ["./packages/kit/src"],
```
grabbou, line 11: by default, `ui/kit` is a server package and to load it on the client, you need to import all packages from the `runtime`.

packages/kit/src/index.tsx
```diff
@@ -1,37 +1,83 @@
...
+})
+
+export type SortableListProps = z.infer<typeof SortableListSchema>
+export const SortableListSchema = z.object({
+  footer: z.any(),
+  items: z.array(SortableListItemSchema),
+  // tbd: support this event again
+  onReorder: z.function().optional(),
+})
+export const SortableList = createServerComponent('RiseSortableList', SortableListSchema.parse)
```
grabbou, line 75: A bit of boilerplate, but it a) works and b) does not leak client code onto the server. Suggestions welcome to reduce it even further down. This needs to be improved, so people can write their own components within a single file, not two or three.

packages/kit/src/index.tsx
```diff
@@ -1,37 +1,83 @@
...
-  },
-  RiseSwitchField: {
-    component: SwitchField,
-    validator: SwitchField.validate,
-  },
+import { ReactElement } from 'react'
+import z from 'zod'
+
+function createServerComponent<T = any>(type: string, validator: (props: T) => T) {
+  return (props: T): ReactElement => ({
```
grabbou, line 5: We can't change return type of JSX Factory, so we have to return shape that satisfies React.Element type. This is an open issue in TypeScript from 2018. This shape is quite good, but we're passing `key: null` since we're not using that field.

packages/react/src/jsx.ts
```diff
@@ -0,0 +1,36 @@
...
+    Object.entries(props).map(([key, value]) => {
+      // tbd: create a global handler registry and register this callback
+      // make sure to clean-up when no longer needed
+      if (typeof value === 'function') {
+        const uuid = 'generate-uuid-for-handler'
+        return [
+          key,
+          {
+            $: 'event',
+            action: uuid,
```
grabbou, line 22: I know we discussed about using paths here, but we don't know the path here as we're not rendering on the server. Theoretically, we could compute path on the server and pass complete pre-rendered tree to the client, but it would mean JSX is now required and outputting a custom JSON (different language, or manually) becomes a challenge. It would also be tricky when refs are used to cross-reference a component or something or there is layout mutation on the client and path is no longer valid.

grabbou: UUID of some sort should be good enough to register and then, de-register (after state update, check if UUID is still present. If not, perform a clean-up)

tsconfig.json
grabbou: That configuration is only specific to our dev setup. In production, developers will be automatically getting this via "exports" field.

grabbou:
closing as we have implemented a lot already. submitting new pr.

PR #21, opened 2024-04-21
feat: parse props after render, to not leak impl. detail into schema validation

grabbou:
You can just require a function to be passed to your component, which greatly simplifies the prop handling. It is going to be easier to share with the server (see #20) and will not require any special schema to validate an event handler.

By "parsing after render", we mean that all internal representations such as $: component or $: event are replaced with React equivalents.

It also means that if you're accepting components as prop, you can simply require a React Component type, not "any" or { $: 'component' } object.


packages/kit/src/SelectField.tsx
```diff
@@ -10,7 +9,7 @@ const SelectFieldProps = z.object({
   id: z.string().optional(),
   label: z.string().optional(),
   unselectedLabel: z.string().optional().default('...'),
-  onValueChange: EventDataStateSchema.optional(),
+  onValueChange: z.function().args(z.string()).optional(),
```
ericvicenti, line 12: can zod validate fn arguments? how does that work? I'm guessing this is just for TS annotation?

ericvicenti (approved):
yes this looks like an improvement, thank you!

packages/kit/src/SelectField.tsx
grabbou: yeah, TS annotation + if I were to guess, typeof check. Nothing more. And there's no way to check for a React component type, but you can use something such as `z.custom(ReactNode)`, which offers TS, but no validation.

PR #22, opened 2024-04-21
fix: sortable list fails to render when header/footer is passed

grabbou:


example/server/state-schema.ts
```diff
@@ -132,7 +133,6 @@ const layerSchema: z.ZodType<Layer> = z.object({
 
 export type LayersMedia = {
   type: 'layers'
-  name?: string
```
grabbou, line 135: small mistake from previous PR. `name?:` should be on `Layer`. I initially did it the other way while I was understanding the way it works.

PR #24, opened 2024-04-22
feat: async events

grabbou:
WIP: Haven't tested / run this yet. Putting out there for initial review / comments.

**This feature is additive - you can rely on the old system if you don't want to use this one.**

There is ActionEventDataState (`{ $: 'event', action: ['navigate'] }`), HandlerEventDataState (created when a function was passed as a prop). HandlerEventDataState can also be "Async". In this case, it means client is awaiting response and backend should notify.

That said, TemplateEvent can be either HandlerEvent (when containing HandlerEventDataState) or ActionEvent (when containing ActionEventDataState). 

WebSocket server allows you to subscribe to just ActionEvents. Everything else gets handled internally. That said, `subscribeEvent` was renamed to `onActionEvent` to avoid confusion. Name is obviously open to further tweaks.

This code allows to write: `await onSubmit()` inside a React component on the client and receive errors / response from the server that is important to the UX. It can also be used to tweak behavior of the Rise mobile app in the future, similar to how you can navigate with `navigate` action. 

Example, you could return "cache: refresh" or similar response to perform cache reload after form is submitted. Some of these may be added automatically in the future for better UX, when e.g. `formHandler()` function is used instead of `handler()`. This needs further verification whether it makes sense.

**Code sample:**

```tsx
{
  $: 'component',
  onPress: handler(() => { console.log('a') })
}
```

Once JSX support is added, this would happen automatically inside JSX factory. Until that happens, users have to wrap their functions with a `handler` or `asyncHandler` depending on whether they want the client to wait or not.

We use this function to convert functions into HandlerEventDataState without having to iterate over entire tree. Instead, we do it at the definition time. However, we may want to do it anyway, because otherwise it's tricky to de-register handlers when element goes out of screen?


packages/ws-client/src/data-source.ts
```diff
@@ -22,14 +28,21 @@ export type UpdateWebsocketMessage = {
   val: JSONValue
 }
 
+export type EventResponseWebsocketMessage = {
+  $: 'evt-res'
+  key: string
+  val: JSONValue
```
ericvicenti, line 34: Lets maybe have one value for resolved and a different value for rejected/caught

packages/ws-client/src/data-source.ts
```diff
@@ -55,14 +68,28 @@ export function createWSDataSource(wsUrl: string): DataSource {
...
+          handlers.forEach((handle) => handle())
+        }
+        break
+      }
+      case 'evt-res': {
+        const resolve = promises.get(event.key)
+        if (resolve) {
+          resolve(event.val)
+          promises.delete(event.key)
+        } else {
```
ericvicenti, line 85: handle reject here, I suppose

packages/ws-client/src/data-source.ts
```diff
@@ -112,8 +141,14 @@ export function createWSDataSource(wsUrl: string): DataSource {
       stores.set(key, newStore)
       return newStore
     },
-    sendEvent: (event) => {
+    sendEvent: async (event) => {
       send({ $: 'evt', event })
+      if (isHandlerEvent(event)) {
+        return new Promise((resolve) => {
+          // tbd: do we want to register a timeout here?
```
ericvicenti, line 148: yes I think we should have some timeout, if it is configurable. Maybe the event definition can have an optional timeout

packages/react/src/events.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { AsyncHandlerEventDataState, HandlerEventDataState, TemplateEvent } from './template'
+
+type Handler = ((payload: any) => any) | ((payload: any) => Promise<any>)
+
+const handlers = new Map<string, Handler>()
+
+export function handler(func: Handler): AsyncHandlerEventDataState | HandlerEventDataState {
+  const key = crypto.randomUUID()
+  handlers.set(key, func)
+  switch (func.constructor.name) {
```
ericvicenti, line 13: this is kind of a whacky way to do it.. although I can see how it is convenient

one concern is that the name may change if the code gets compressed by a build step somewhere

ericvicenti (changes requested):
Good progress!

packages/ws-client/src/data-source.ts
grabbou: Could be, tricky to parse otherwise!

packages/react/src/events.tsx
grabbou: Yeah, or it gets transpiled to a non-async environment, in that case it's all functions (or sometimes even generators). 

I think once we have JSX support, we should have two ways:
- pure function is passed `prop={func}` - no response to the server
- wrapped with handler - `prop={handler(func)}` - response to the server

I don't think there's other reliable way.

For now, we need to use function in both cases as we don't have JSX sugar, so I guess calling `handler` or `asyncHandler` depending on your use case is going to be easiest.

packages/react/src/template.tsx
```diff
@@ -32,16 +32,20 @@ export type AsyncHandlerEventDataState = {
...
   async: true
+  // server-only
+  handler?: EventHandler
 }
 export type HandlerEventDataState = {
   $: 'event'
   key: string
   async: false
+  // server-only
+  handler?: EventHandler
```
ericvicenti: hmm.. not super happy about this, but I'm glad it works

we could alternatively have a higher level server state that has `EventHandlerState & { handler: ...}`

grabbou: yeah, this is types thing. I will have to rework it, but otherwise I think it's a good way.

packages/ws-client/src/data-source.ts
grabbou: yes, let's allow optional timeout to overwrite on per event basis. What should be the default tho?

grabbou:
@ericvicenti let me know if there's anything suspicious apart from already left comments. Will address them later today/tomorrow.

packages/react/src/refs.tsx
```diff
@@ -163,22 +163,29 @@ function resolveRef(dataValues: DataValues, lookup: ReferencedDataState['ref']):
 export function Template({
   components,
   dataSource,
-  path = '',
+  storeKey = '',
   onEvent,
 }: {
-  path?: ReferencedDataState['ref']
+  storeKey?: string
```
grabbou, line 169: tbd: `storeKey` will be incorrect if there's another component referenced. If you ref component from another store, the storeKey should be of that store, otherwise it's not going to work. Will fix later.

grabbou:
Once #25 is merged into #24, this PR should automatically become green and be mergeable.

packages/ws-client/src/data-source.ts
grabbou: will do a timeout in a follow-up + some unit tests

grabbou:
Ready for another pass! It's non-breaking PR. I plan to use async event handling code when building demo server for the App Store.

packages/react/src/template.tsx
```diff
@@ -137,56 +160,47 @@ export function BaseTemplate({
...
-            action: node.action,
-            payload,
-          })
+    if (isEventDataState(propValue)) {
+      return async (payload: any) => {
+        // React events (e.g. from onPress) contain cyclic structures that can't be serialized
+        // with JSON.stringify and also provide little to no value for the server.
+        // tbd: figure a better way to handle this in a cross-platform way
+        if (payload?.nativeEvent) {
+          payload = '[native code]'
```
ericvicenti, line 173: hmm yeah this isn't ideal. But we can figure out a way to improve it later

ericvicenti (approved):
Good stuff, can't wait to see this async event behavior in "real world" usage!

packages/react/src/template.tsx
grabbou: at least we're temporarily as fancy as dev tools console.log 😆

PR #27, opened 2024-04-24
feat(eg): Library saving, value smoothing, UI improvements

ericvicenti:


grabbou:
Rebased after recent changes and applied necessary tweaks.

PR #29, opened 2024-04-25
feat: add example server

grabbou:
Building a very simple inventory server that we will use as one of the defaults to showcase the library. There's still a lot of higher-order functionalities missing, particularly around state update, routing, so the current code may look a bit rough at first sight.

We will use this PR and others as an opportunity to further improve the high-level API.

The details screen is pending and will be added in a follow-up. It will include a big picture of the item, title and a simple form to change price/quantity.


example/demo/src/inventory/ui.ts
```diff
@@ -0,0 +1,133 @@
...
+          (item, idx): ServerDataState => ({
+            $: 'component',
+            key: item.key,
+            component: 'Button',
+            props: {
+              unstyled: true,
+              // onPress: {
+              //   $: 'event',
+              //   action: ['navigate', `inventory:item:${item.key}`],
+              // },
```
grabbou, line 41: Will be added in a follow-up PR together with some nice forms.

grabbou:
Details screen will allow to:
- browse detailed description of an item
- change quantity with a +/- button
- update price
- we will take advantage of async functionality we have

packages/playground/src/screens/home.tsx
```diff
@@ -5,29 +5,30 @@ import { ImageURISource, ScrollView } from 'react-native'
...
+            {state.connections.map((connection) => (
+              <ConnectionItem key={connection.id} connection={connection} />
+            ))}
+          </YGroup>
+          <NewConnectionButton />
+        </YStack>
+        <YStack gap="$2">
+          <Text textAlign="center">or try some of the examples below:</Text>
+          <YGroup separator={<Separator />}>
+            {Object.values(BUILTIN_CONNECTIONS).map((connection) => (
```
ericvicenti, line 27: I think we should hide these once you have added your own connection

ericvicenti (approved):
Great!!

packages/playground/src/screens/home.tsx
grabbou: Maybe swap with a button so you can still play around? Otherwise won't ever be able to check them out!

ericvicenti: Sure, that could work. But once you have added your own server you probably don't really care about the examples. And you could always follow a URL to open them from the website

PR #32, opened 2024-04-26
feat: do not render default connections when custom are set

grabbou:
As per discussion in the last PR, we don't want to render example connections when custom were provided by the user.


grabbou:
Since it's just tests, I am going to go ahead and merge it to clean things up!

PR #33, opened 2024-04-26
wip: status indicator and error component per data source

grabbou:
Work in progress. Trying to figure best way to integrate status indicator and error component that is scoped to Data Source. I think that is most universal, since each data source will have different transportation layer and may need to use different logic to display its status or provide dev tools support.


ericvicenti (changes requested):
Continuing the discussion on discord, I don't think the data source should have any views or dependency on Tamagui

grabbou:
Yes, to sum up:
- use streams and stream updates from data source
- create separate component to provide necessary UI

PR #34, opened 2024-04-27
feat: add ability for DataSource to stream updates

grabbou:
Apart from comments, here is a question:

If we provide "DataBoundary" as a component to be rendered as follows:
```tsx
<DataBoundary>
   <Template />
</DataBoundary>
```

We can only render overlays (such as status indicator) in context of that screen, unless we use portals. I think that is ultimately a good thing, because we're not interfering with a header (that may be hidden, or have some other widgets in it).

If we agree that this is good direction, then, another question would be how to render debugging information, such as "status indicator"?

Here are some suggestions to kick start the discussion:

- Extend dev tools menu with "Toggle Final UI monitor" (https://docs.expo.dev/develop/development-builds/development-workflows/#extending-the-dev-menu). This would be living in Rise and would require users to do additional step on their end to enable this too. When toggled, display a position: fixed overlay in the bottom corner showing status and some other performance numbers.
- Do not render status indicator at all. Assume it's always connected. If it disconnects, use a "You are offline" horizontal bar that we attach to the top of the screen together with "Tap to reconnect" button. Or use a toast instead. If there's no UI to render, use a full-screen "You are offline" message instead.

Examples:
- toast 
- horizontal bar attached to the header


packages/playground/src/data-boundary.tsx
```diff
@@ -0,0 +1,13 @@
+import { DataSource, useStream } from '@final-ui/react'
+import React, { PropsWithChildren } from 'react'
+import { Text, YStack } from 'tamagui'
+
+export function DataBoundary({ info, children }: PropsWithChildren<{ info: DataSource['info'] }>) {
```
grabbou, line 5: This package should live in core and be either provided out of the box or already wrapping Template.

packages/react/src/refs.tsx
```diff
@@ -11,13 +12,15 @@ import {
   TemplateEvent,
 } from './template'
 
-export type Store<V = JSONValue> = {
-  get: () => V
-  subscribe: (handler: () => void) => () => void
+type DataSourceInfo = {
+  status: 'connecting' | 'connected' | 'closed' | 'closing' | 'error' | 'undetermined'
```
grabbou, line 16: RWS does not report "connecting" and "closing" statuses, they're not part of the spec. Do we still want to keep them? (I suggest let's not)

packages/react/src/refs.tsx
```diff
@@ -11,13 +12,15 @@ import {
...
 }
 
+export type Store<T = JSONValue> = Stream<T>
+
 export type DataSource = {
   get: (key: string) => Store
+  info: Stream<DataSourceInfo>
```
grabbou, line 23: Name suggestions welcome

packages/react/src/template.tsx
```diff
@@ -200,30 +200,7 @@ export function BaseTemplate({
     return propValue
   }
 
-  return <ErrorBoundary>{render(dataState, path)}</ErrorBoundary>
```
grabbou, line 203: This error boundary will be implemented inside DataBoundary instead.

packages/ws-client/src/data-source.ts
```diff
@@ -42,7 +43,7 @@ export type ClientWebsocketMessage =
 
 export type ServerWebsocketMessage = UpdateWebsocketMessage | EventResponseWebsocketMessage
 
-type Handler = () => void
+type Handler = (value: JSONValue) => void
```
grabbou, line 46: Aligned "createStore" with "Store" type which is now a Stream. We have a few different variants of get/subscribe throughout the codebase, so I was trying to unify in a way that they follow same shape.

packages/ws-client/src/data-source.ts
```diff
@@ -160,3 +162,29 @@ export function createWSDataSource(wsUrl: string): DataSource {
...
+
+  const updateStatus = () => writeInfo({ status: readyStateToStatus(rws.readyState) })
+  rws.addEventListener('open', updateStatus)
+  rws.addEventListener('close', updateStatus)
+  rws.addEventListener('error', updateStatus)
+
+  return info
+}
+
+const readyStateToStatus = (readyState: number) => {
```
grabbou, line 177: RWS does not report "connecting" and "closing" statuses, but they are valid as per spec. I think they may happen inside `onError()` event. I always read the status from `rws.readyState` and convert to one of these.

packages/playground/src/data-boundary.tsx
grabbou: I would keep the header free from any status indicators, because users may utilise header buttons with their own UI. I suggest we come up with a nice non-intrusive UI to display this, maybe similar to React Frames Count overlay.

grabbou:
Update

Simulator Screenshot - iPhone 15 Pro - 2024-04-30 at 14 43 35
simulator_screenshot_786A0412-49A8-4CBD-9709-55FDBF2E7ACE

packages/playground/src/screens/connection.tsx
```diff
@@ -43,6 +44,12 @@ function ActiveConnectionScreen({ connection }: { connection: Connection }) {
     return null
   }
 
+  useEffect(() => {
```
grabbou, line 47: Primarily during development, path occasionally becomes undefined. This is probably an issue with fast refresh or something. Anyway, this makes sure that whenever that happens, we just go back to the Home Screen.

ericvicenti: ah interesting! I will keep an eye out for unexpected navigation back

packages/react/src/refs.tsx
```diff
@@ -116,8 +114,8 @@ function createRefStateManager(
   function ensureSubscription(key: string) {
     if (!refSubscriptions[key]) {
       const refValueProvider = dataSource.get(key)
-      refSubscriptions[key] = refValueProvider.subscribe(() => {
-        const didUpdate = setRefValue(key, refValueProvider.get())
+      refSubscriptions[key] = refValueProvider.subscribe((value) => {
```
ericvicenti, line 117: btw for subscriptions I followed the API provided by React useSyncExternalStore. where the event does not contain the data and instead it will call `get` again

https://react.dev/reference/react/useSyncExternalStore

I used to have useSyncExternalStore but our stuff is too complicated 😂

packages/react/src/streams.ts
```diff
@@ -0,0 +1,47 @@
+import { useSyncExternalStore } from 'react'
+
+export type Stream<V> = {
+  get: () => V
+  subscribe: (handler: (value: V) => void) => () => void
+}
+
+export type DataSourceStateStream = Stream<{
+  status: 'connecting' | 'connected' | 'closed' | 'closing' | 'error' | 'undetermined'
```
ericvicenti, line 9: These are fine for now but not great. What if you are `closed` but you also have an error? Also, what is the purpose of `undetermined`? Does that basically mean "closed"? Eventually maybe error should be in a separate field

Also the opposite of "connected" is "disconnected", not "closed"

Also I'm not sure "closing" will ever be used. You should be able to instantly close the connection from the client

packages/ws-client/src/data-source.ts
ericvicenti: sweet!

packages/ws-client/src/data-source.ts
```diff
@@ -139,7 +145,7 @@ export function createWSDataSource(wsUrl: string): DataSource {
 
   const promises = new Map<string, [PromiseHandler, PromiseHandler]>()
 
-  const dataSource: DataSource = {
```
ericvicenti, line 142: one reason I prefer to do it this way is that the type error will appear in a more helpful place when you try to assign it to the `DataSource` type immediately

packages/ws-client/src/data-source.ts
ericvicenti: Ok nice

packages/playground/src/data-boundary.tsx
```diff
@@ -0,0 +1,44 @@
...
+import React, { PropsWithChildren } from 'react'
+import { H4, SizableText, Text, YStack } from 'tamagui'
+
+// tbd: convert `WebSocketDataSource` to `NetworkDataSource`
+export function DataBoundary({
+  dataSource,
+  path,
+  children,
+}: PropsWithChildren<{ dataSource: WebSocketDataSource; path: string }>) {
+  const { status } = useStream(dataSource.state)
```
ericvicenti, line 13: as discussed, we eventually want to remove the state from the dataSource, and allow the user to pass in the state stream directly into DataBoundary

ericvicenti (approved):
Overall, good change!

packages/playground/src/data-boundary.tsx
grabbou: Yeah, I think as a mid-step we can make "state" stream a separate prop, to make it more explicit that it can come from a different place. 

However, I think for "WebSocketDataSource" or other DataSources that are network-related, it is okay if they have an extra "state" prop that is not included in the base DataSource type. 

This is what we have right now.

packages/ws-client/src/data-source.ts
grabbou: Correct! Will revert back this change, you're right!

packages/react/src/streams.ts
grabbou: Yeah,

"undetermined" is carry over from previous work. I would remove it, as it can't happen in the WebSocket connection.

All other statuses are directly taken from WebSocket. However, I think to further simplify this, we should just make it "connected" and "disconnected" (or "closed", to unify with WebSocket spec), because these are the only statuses that we can listen to (via "connected" and "closed" events). All other statuses, while being part of the spec, are internal and we can't really get them.

"error" should also be removed. You can have an error and still be connected, but in case it throws, it will just change status to disconnected.

WDYT?

ericvicenti: Yep! Error could be a different stream

packages/react/src/refs.tsx
grabbou: Haha! Which one you prefer tho? Our "streams" implementation passes "value" to the handler(), so our lambda passed to subscribe() has access to "value". I guess it boils down to preference, and I am fine with doing it either way!

packages/playground/src/data-boundary.tsx
ericvicenti: Yep totally fine if some data sources have extra features like the status. I just want to remove it from the main DataSource type

grabbou: Yes, it's not included. This PR, if you look at WebSocket data source, has a different type:
```ts
export type WebSocketDataSource = DataSource & {
  state: DataSourceStateStream
}
```

grabbou: In the future, this will be probably named a higher-level "NetworkDataSource" which is a DataSource with "state". But too soon to speculate that.

packages/react/src/streams.ts
grabbou: Simplified to just "connected" and "disconnected". It can also be undefined, which means not connected and not disconnected (default state).

packages/ws-client/src/data-source.ts
grabbou: Going to leave it as is for now, but feel free to change later. From the screenshot attached, both versions seem to provide same level of DX.

Screenshot 2024-05-01 at 13 25 46
Screenshot 2024-05-01 at 13 25 19

PR #36, opened 2024-04-29
feat: migrate demo server to Cloudflare Workers and setup deployment pipeline

grabbou:
Convert server to Cloudflare Worker
Update env variables
Add a GitHub Action to trigger deployment (maybe we can have deployments from `main` once work stabilises)


packages/ws-server/src/ws-rnt-server.ts
```diff
@@ -23,8 +23,16 @@ type ActionEventHandler = (
...
+  const dataSource = createWSDataSource()
+  const wss = new WS.Server({ port })
+  wss.on('listening', () => {
+    console.log(`WebSocket server started on port ${port}`)
+  })
+  wss.on('connection', dataSource.handleWSConnection)
+  return dataSource
+}
 
+export function createWSDataSource() {
```
grabbou, line 35: I made this function WebSocket.Server agnostic. It just returns `handleWSConnection` function that will take any HTML5 compatible WebSocket instance. Works with "ws" package (Node) as well as Cloudflare WebSocket.

Check "example/demo/src/index.ts" for usage:
```ts
const webSocketPair = new WebSocketPair()
		const [client, server] = Object.values(webSocketPair) as [WebSocket, WebSocket]

		// Accepts the WebSocket connection and begins terminating requests
		// for the WebSocket on Cloudflare’s global network.
		server.accept()

		const dataSource = createWSDataSource()
		dataSource.handleWSConnection(server)
```

or 

```ts
const dataSource = createWSDataSource()
  const wss = new WS.Server({ port })
  wss.on('listening', () => {
    console.log(`WebSocket server started on port ${port}`)
  })
  wss.on('connection', dataSource.handleWSConnection)
  return dataSource
```

- `createWSServer` - creates an entire WebSocket server for you. I believe this should ultimately go to a separate package.
- `createWSDataSource` - creates a data source that works with any WebSocket connection

**Now, the problem is that "ws-client" package also has "createWSDataSource" function, so we have same name and it gets confusing (or not?)**

I am not sure what's the best way to name all of the following:
- `ws-client` - `createWSDataSource()`
- `ws-server` - `createWSServer()`
- `ws-server` - `createWSDataSource()`

grabbou: Maybe it all should go into single package, called "ws" and do not depend on anything Node/Web specific. Also, since both `ws-server` "handleWSConnection()" and `ws-client` packages work with "WebSocket" HTML5 instance, it means that you can technically run both on server/client (wherever you want).

So maybe the names should be "WSReceiver" and "WSSender" or something.

.github/workflows/deploy-demo-server.yml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          cache: 'npm'
+
+      - run: npm install
+
+      - name: Deploy
+        uses: cloudflare/wrangler-action@v3
+        with:
+          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          workingDirectory: "./example/demo"
```
grabbou, line 26: Hard to test this in a PR, but should work. If it doesn't, I will send a fix directly to master in a follow-up.

packages/ws-server/src/ws-rnt-server.ts
ericvicenti: I think we can support CF workers without having this extra step. So that the default module uses regular WS server, and if you have CF you can do something like this, sorry I don't actually know the full CF WS API

```ts
createWSDataSource({
  createServer: (...something) => {
    const webSocketPair = new WebSocketPair()
    const [client, server] = Object.values(webSocketPair) as [WebSocket, WebSocket]
    server.accept()
    // use ...something
  }
})
```

grabbou: Sticking to `createWSServerDataSource` for now as per our discussion

PR #41, opened 2024-05-01
feat: experimental JSX support

grabbou:
Ready.


example/demo/tsconfig.json
```diff
@@ -5,10 +5,14 @@
 		"lib": ["es2021"],
 		"module": "es2022",
 		"moduleResolution": "Bundler",
-		"types": ["@cloudflare/workers-types"]
+		"types": ["@cloudflare/workers-types"],
+		"jsx": "react-jsx",
+    "jsxImportSource": "@final-ui/react"
```
grabbou, line 10: All that is needed to enable custom JSX. We would ship a preset so this does not need to be set by users.

example/demo/src/inventory/ui.tsx
```diff
@@ -173,142 +175,91 @@ export function getHomeScreen(): ServerDataState {
...
 }
 
+// tbd: we need a better way to design this API, should ideally come from tamagui package.
+const YStack = createComponentDefinition<typeof t.YStack>('YStack')
+const XStack = createComponentDefinition<typeof t.XStack>('XStack')
+const Image = createComponentDefinition<typeof t.Image>('Image')
+const H2 = createComponentDefinition<typeof t.H2>('H2')
+const SizableText = createComponentDefinition<typeof t.H2>('SizableText')
+const Paragraph = createComponentDefinition<typeof t.H2>('Paragraph')
+const Button = createComponentDefinition<typeof t.H2>('Button')
```
grabbou, line 185: I am not sure what is the best way to declare importable components that can be used on the Server, ideally, without having to write this.

I was experimenting with the following:
- import any component on the server (e.g. import { YStack } from "tamagui")
- if "component function" has "validator" property, use it on the client to validate props
- on the client, do "import * as TamaguiComponents from "tamagui"" and pass that to `<Template components={TamaguiComponents} />"

This would allow fully seamless code sharing, with one (possibly deal breaking) caveat - if user minifies code at some point down the road, and "import * as TamaguiComponents from "tamagui"" has different names in the client bundle (e.g. "YStack" is "Y$"), the server would still output "YStack" and fail.

Unfortunately, it would require server code to depend on React / Tamagui, which is not ideal.

example/demo/src/inventory/ui.tsx
```diff
@@ -173,142 +175,91 @@ export function getHomeScreen(): ServerDataState {
...
+  // @ts-ignore return type of JSX factory needs to be changed
+  return (
+    <YStack flex={1} backgroundColor={'$background'} gap="$3">
+      <Image
+        key="photo"
+        // @ts-ignore ref needs to be valid prop
+        source={{
+          uri: {
+            $: 'ref',
+            ref: [`inventory-items`, item.key, 'photo'],
```
grabbou, line 197: Need to do some tweaks to the JSX types to allow passing "ref"s.

example/demo/src/inventory/ui.tsx
```diff
@@ -173,142 +175,91 @@ export function getHomeScreen(): ServerDataState {
...
+          }}
+        />
+        <XStack key="adjustments" gap="$3" alignItems="center">
+          <XStack key="quantity" gap="$2">
+            <SizableText size="$5">Quantity:</SizableText>
+            <SizableText
+              size="$5"
+              children={{
+                $: 'ref',
+                ref: [`inventory-items`, item.key, 'quantity'],
```
grabbou, line 229: We may need a <Ref /> component here so it can be passed more common way, rather than object.

example/demo/src/inventory/ui.tsx
ericvicenti: Can we use a different import?

Client:
```
import { RiseComponents } from "@final-ui/kit"
<Template components={RiseComponents} />
```

```
import { RiseSwitchField } from "@final-ui/kit/components"

// write on the server
<RiseSwitchField />
```

example/demo/tsconfig.json
ericvicenti: Ok I guess this means that users cannot use JSX if their server code also uses React SSR.

Shouldn't be a big deal but it does seem unfortunate

example/demo/src/inventory/ui.tsx
```diff
@@ -1,4 +1,6 @@
 import { handler, ServerDataState } from '@final-ui/react'
+import { createComponentDefinition } from '@final-ui/react/jsx-runtime'
+import type * as t from 'tamagui'
```
ericvicenti, line 3: why importing from Tamagui instead of importing from Final Kit?

packages/react/src/jsx.tsx
```diff
@@ -0,0 +1,51 @@
...
+  key: string
+): ComponentDataState {
+  const { type, props } = componentFactory(passedProps)
+  if (typeof type !== 'string') {
+    throw new Error('Invalid component.')
+  }
+  const serialisedProps = Object.fromEntries(
+    Object.entries(props).map(([key, value]) => {
+      if (typeof value === 'function') {
+        // tbd: how do we now if it's async or not?
```
ericvicenti, line 25: I wonder, do we need to know if it is async? There should be a way to handle it transparently

On the client we will always have an async function, but on the server it shouldn't matter.. right?

ericvicenti (approved):
good progress!

example/demo/tsconfig.json
grabbou: Yeah, or... we could use Babel transformer with conditionals (or custom file extension). I guess we can provide this in "advanced" section of some guides we make.

example/demo/src/inventory/ui.tsx
grabbou: Yes, right now, in this PR, we import from the "kit" or "kit/runtime" (on the client). So, your suggestion is pretty much the same, but reverse, which is also fine.

We should however discuss how to provide those wrapper components for the server, not sure what's the best way to do it.

packages/react/src/jsx.tsx
grabbou: So this comment means that if it's async handler, we should wrap it with `asyncHandler`. This registers promise on the client with id, that will be resolved when server sends response back to the client.

In the other PR, we discussed that we would go with async-only behavior, and I mentioned it would simplify stuff once JSX support is in. 

This is the place that would be simplified. We just treat all as async anyway.

example/demo/src/inventory/ui.tsx
ericvicenti: Well, the issue is the dependency on `@final-ui/kit` from the server package? Can't we export the actual typed JSX components from the `kit` package?

packages/kit/package.json
```diff
@@ -3,7 +3,10 @@
   "version": "0.2.0",
   "license": "Apache-2.0",
   "description": "",
-  "main": "lib",
+  "exports": {
+    ".": "./lib/index.js",
+    "./server": "./lib/server.js"
```
grabbou, line 8: tbd: support experimental React Native exports too.

apps/mobile/tsconfig.json
```diff
@@ -1,5 +1,8 @@
 {
   "extends": "../../tsconfig",
+  "compilerOptions": {
+    "baseUrl": "../.."
+  },
```
grabbou, line 5: We set this so that Expo / TypeScript knows to resolve paths (set in root tsconfig.json) from root of this workspace.

tsconfig.json
```diff
@@ -2,8 +2,10 @@
   "extends": "@tsconfig/node18",
   "compilerOptions": {
     "paths": {
-      "@final-ui/app": ["./packages/playground/src"],
+      "@final-ui/playground": ["./packages/playground/src"],
       "@final-ui/react": ["./packages/react/src"],
+      "@final-ui/react/server": ["./packages/react/src/server"],
+      "@final-ui/react/jsx-runtime": ["./packages/react/src/jsx"],
       "@final-ui/ws-client": ["./packages/ws-client/src"],
```
grabbou, line 9: In production, we use `lib`. In development, we use `src` and compile with TypeScript to save time (no need to run `watch` and re-compile `lib`). Added two more paths as per "exports" field and fixed incorrect package name.

packages/react/src/jsx.tsx
```diff
@@ -0,0 +1,48 @@
...
+  }
+}
+
+export function createComponentDefinition<
+  T extends JSXElementConstructor<any> | keyof JSX.IntrinsicElements,
+  P = React.ComponentProps<T>,
+>(type: string) {
+  return (props: P): ReactElement => ({
+    type,
+    props,
```
grabbou, line 45: Right now, TypeScript does not allow to change the type of JSX factory to something non-standard. To avoid type errors, we have to return "ReactElement". So, `type`, `props`, and `key`. We don't need key (yet), and also, it's computed on the client.

packages/react/src/jsx.tsx
```diff
@@ -0,0 +1,48 @@
...
+  { children, ...passedProps }: Props,
+  key: string
+): ComponentDataState {
+  const { type, props } = componentFactory(passedProps)
+  if (typeof type !== 'string') {
+    throw new Error('Invalid component. Make sure to use server-side version of your components.')
+  }
+  const serialisedProps = Object.fromEntries(
+    Object.entries(props).map(([key, value]) => {
+      if (typeof value === 'function') {
```
grabbou, line 24: Here, we automatically wrap functions (e.g. onPress={() => {}}) but we should also automatically wrap objects that are ActionDataState or an array of them.

example/demo/tsconfig.json
grabbou, line 10:
```
"jsx": "react-jsx",
    "jsxImportSource": "@final-ui/react"
```
is all you have to set to enable JSX support. Will be currently manual, but in next PR we should create a preset.

ericvicenti: Right, makes sense

apps/mobile/tsconfig.json
ericvicenti: ok gotcha

tsconfig.json
ericvicenti: nice this sounds good

packages/react/src/jsx.tsx
ericvicenti: Ah. Well, we do need key, right? It may be necessary for some devs to re-order the elements of an array, and key is the only way

packages/react/src/jsx.tsx
ericvicenti: right that makes sense

packages/react/src/jsx.tsx
grabbou: Yes - correct. If you look at `jsx` function factory, we get `key` and pass it to the client, so you can set keys. This is just a "componentDefinition" function that allows to write <Component /> on the server. 

```ts
export function jsx(
  componentFactory,
  { children, ...passedProps },
  key,
): ComponentDataState {
  const { type, props } = componentFactory(passedProps)
  return {
      $: 'component',
      component: type,
      key,
      props,
      children,
   }
}
```  

So, here's few things happening:
- `componentFactory` takes `props` and returns `props`. This is not really necessary - because we use `componentFactory` really to just get `type` of component (its stringified name). I left it this way in case we wanted to add validation on the server-side in the future. If yes, that would happen inside `componentFactory`
- we do not pass `key` to component factory, because its purpose is to just get component type and (in the future) props - as per previous point. We still put key on the "ComponentDataState"

ericvicenti: The last bullet is the only one that I don't understand. What is the `componentDefinition` in comparison to the `componentFactory`?

example/demo/src/inventory/ui.tsx
```diff
@@ -0,0 +1,165 @@
...
+} from '@final-ui/tamagui/server'
+
+import { UIContext } from '../types'
+import inventory, { Item } from './inventory'
+
+// Define JSX Element and ElementType
+// Need to unload `React` or remove `React.JSX.Element` and `React.JSX.ElementType` from global namespace
+// to run this
+// WIP: overwrite instead
+declare global {
```
grabbou, line 27: last tweak needed:
1. Figure out how to not load React / overwrite React namespace
2. Move this into d.ts and load automatically
These two types change what JSX returns to ComponentDataState.

packages/react/src/jsx.tsx
grabbou:
```
componentFactory: ReturnType<typeof createComponentDefinition>,
```

PR #46, opened 2024-05-05
feat: ability to respond to events with client-side actions

grabbou:
This PR adds ability for the data source to have impact on client-side behavior after responding to the request. It misses some tests and fine tuning, but I wanted to make sure we have directional agreement before committing more time to it. After initial approval, I will go ahead and complete missing tests.

On a high-level, from your event handler, you can return either plain object or an instance of Response:
```ts
{
  $: 'component',
  component: 'RiseForm',
  props: {
    onSubmit: handler((values) => {
      // handle on press, then
      return res.json()
        // send action
        .action(['navigate', 'goBack'])
        // or more
        .action(['some-other-action'])
    })
  }
}
```

Above example will instruct client to perform "goBack" navigation which is specific to Rise mobile application after 
succesfully handling action on the server. We should document them at some point, most likely in a follow-up PR, as we currently have just two.

> [!NOTE]
> In the future, we can hide `res.json().action(['navigate', 'goBack'])` behind something nicer.

## Other notable changes

- Remove non-async events
- Rename ActionEvent to Action
- Do not send ActionEvents to the server at all
- Updated JSONValue / DataState types. JSONValue is now included in DataState, not the other way around. That way, JSONValue cna be used as a type in places that expect actual JSON values.

### DIFF
--- a/packages/playground/src/screens/connection.tsx
+++ b/packages/playground/src/screens/connection.tsx
@@ -76,13 +76,22 @@ function ActiveConnectionScreen({ connection }: { connection: Connection }) {
         }
       }
       const res = await dataSource.sendEvent(event)
-      // res is `null` if we sent `actionEvent` to the server
-      // we would have to disable sending action events to the server to make response always defined
-      if (res?.actions) {
+      if (!res) {
+        // res is `null` if we sent `actionEvent` to the server
+        // we would have to disable sending action events to the server to make response always defined
+        return
+      }
+      if (res.actions) {
+        // response has local actions to execute on the client as a follow-up
         for (const action of res.actions) {
           handleActionEvent(action)
         }
       }
+      if (!res.ok) {
+        // if response was not okay, throw the payload as an error
+        throw res.payload
+      }
+      // resolve otherwise
       return res?.payload
     },
     [dataSource, handleActionEvent]
--- a/packages/playground/src/screens/connection.tsx
+++ b/packages/playground/src/screens/connection.tsx
@@ -50,7 +50,7 @@ function ActiveConnectionScreen({ connection }: { connection: Connection }) {
     }
   }, [path])
 
-  const handleActionEvent = useCallback(
+  const onAction = useCallback(
     (dataState: ActionEventDataState) => {
       const [action, path] = Array.isArray(dataState.action)
         ? dataState.action
@@ -67,39 +67,9 @@ function ActiveConnectionScreen({ connection }: { connection: Connection }) {
     [router]
   )
 
-  const onEvent = useCallback(
-    async (event: TemplateEvent) => {
-      if (isActionEvent(event)) {
-        if (handleActionEvent(event.dataState)) {
-          // do not send handled events to the server
-          return
-        }
-      }
-      const res = await dataSource.sendEvent(event)
-      if (!res) {
-        // res is `null` if we sent `actionEvent` to the server
-        // we would have to disable sending action events to the server to make response always defined
-        return
-      }
-      if (res.actions) {
-        // response has local actions to execute on the client as a follow-up
-        for (const action of res.actions) {
-          handleActionEvent(action)
-        }
-      }
-      if (!res.ok) {
-        // if response was not okay, throw the payload as an error
-        throw res.payload
-      }
-      // resolve otherwise
-      return res?.payload
-    },
-    [dataSource, handleActionEvent]
-  )
-
   return (
     <DataBoundary dataSource={dataSource} path={path!}>
-      <Template components={components} dataSource={dataSource} path={path!} onEvent={onEvent} />
+      <Template components={components} dataSource={dataSource} path={path!} onAction={onAction} />
     </DataBoundary>
   )
 }
--- a/packages/playground/src/screens/connection.tsx
+++ b/packages/playground/src/screens/connection.tsx
@@ -1,5 +1,5 @@
 import { RiseComponents } from '@final-ui/kit'
-import { Template } from '@final-ui/react'
+import { ActionEventDataState, Template } from '@final-ui/react'
 import { TamaguiComponents } from '@final-ui/tamagui'
 import { Stack } from 'expo-router'
 import React, { useCallback, useEffect } from 'react'
@@ -51,13 +51,13 @@ function ActiveConnectionScreen({ connection }: { connection: Connection }) {
   }, [path])
 
   const onAction = useCallback(
-    (action: string | string[]) => {
-      const [operation, path] = Array.isArray(action) ? action : [action, '']
-      if (operation === 'navigate') {
+    (event: ActionEventDataState<string | string[]>) => {
+      const [action, path] = Array.isArray(event.action) ? event.action : [event.action, '']
+      if (action === 'navigate') {
         router.push(`/connection/${params.id}?path=${path}`)
         return true
       }
-      if (operation === 'navigate-back') {
+      if (action === 'navigate-back') {
         router.back()
         return true
       }
--- a/packages/playground/src/screens/connection.tsx
+++ b/packages/playground/src/screens/connection.tsx
@@ -2,7 +2,7 @@ import { RiseComponents } from '@final-ui/kit'
 import { isActionEvent, Template, TemplateEvent } from '@final-ui/react'
 import { TamaguiComponents } from '@final-ui/tamagui'
 import { Stack } from 'expo-router'
-import React, { useCallback, useEffect } from 'react'
+import React, { useCallback } from 'react'
 import { createParam } from 'solito'
 import { useRouter } from 'solito/router'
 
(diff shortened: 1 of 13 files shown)


grabbou:
Also, going forward, I think we should drop `handler` (the one that does not allow you to return from the server, client does not await response) and just stick to `asyncHandler`. 

This will solve one major problem: after we switch to JSX, having to decide whether the following:
```
<Component onSubmit={() => // do something, return new Error("Failed")} />
```
is a `handler` (no response supported by the client) or `asyncHandler` (you can return response, client will wait if component is able to handle this UX).

By making it just one, we reduce complexity and get rid of challenges like the one above.

In practice, you want to return something to the client most of the time to indicate "success". If you still want to perform a side-effect without user feedback, you can make your handler function return nothing.

This does not have any impact on the performance of the client-side, because no handlers are awaited. It is only custom components in Final UI kit that have this capability and users are expected to do that on their own too. 

Note: you can still do `action: ['navigate-back]'` to perform client-side navigation without bothering the server.

ericvicenti (changes requested):
Very cool progress!

I think its fine to support async-only server behaviors.

I'm really not sure how much of this stuff should be strictly tied to every data source, we should discuss this

packages/react/src/response.ts
```diff
@@ -0,0 +1,42 @@
...
+// an appropriate error state
+export const res = {
+  json(payload: JSONValue) {
+    const res = new ServerResponse()
+    res.payload = payload
+    return res
+  },
+}
+
+export class ServerResponse {
```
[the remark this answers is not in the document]
grabbou: Yeah, we can do it this way. I wasn't really sure of the best pattern, so did a quick POC to discuss!

packages/playground/src/screens/connection.tsx
```diff
@@ -44,35 +44,36 @@ function ActiveConnectionScreen({ connection }: { connection: Connection }) {
     return null
   }
 
+  useEffect(() => {
```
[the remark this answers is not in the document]
grabbou: > Also it seems possible that the events would be fired multiple times

Yeah, this is possible. In which case, it is theoretically possible a navigation would happen twice, if you render a Provider component more than once.

However, in practice, we would have Rise-specific (because those navigation events are exclusive to Rise) "NavigationProvider" that would be responsible to handle navigation-only events.

We would then, create providers / specialised components for each type of events (e.g. "analytics" or "logging"), making the design overall modular.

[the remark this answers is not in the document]
grabbou: > the effect is not as clean as the previous implementation, in my opinion

We could theoretically keep it within `<Template />` component, but I consider this an unnecessary API since `onEvent` is not tied to Template component, but to DataSource (in this particular PR).

Happy to discuss, also, maybe we can come up with a nice hook such as `useTemplateEvent(dataSource)` or something.

packages/react/src/template.tsx
```diff
@@ -27,7 +33,7 @@ export type ReferencedDataState = {
   ref: Path
 }
 type EventDataState = ActionEventDataState | HandlerEventDataState
-export type ActionEventDataState<T = any> = {
+export type ActionEventDataState<T = string | string[]> = {
```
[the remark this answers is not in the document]
grabbou: Yeah, should be opaque. This type annotation just changes the default value. However, since it's opaque, we shouldn't probably make assumptions about its type! Will double check this.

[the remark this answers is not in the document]
ericvicenti: Oh yeah hmm, probably not a big deal. For some reason I misread this as `extends`

packages/playground/src/screens/connection.tsx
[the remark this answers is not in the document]
grabbou: As discussed, we will be returning back to previous implementation to avoid combining data source with client-side only events.

packages/react/src/refs.tsx
```diff
@@ -160,15 +161,20 @@ function resolveValueRefs(dataValues: DataValues, value: JSONValue): JSONValue {
     if (isCompositeDataState(value) && value.$ === 'ref') {
       return resolveRef(dataValues, value.ref)
     }
+    // @ts-ignore
     return Object.fromEntries(
```
grabbou, line 165: please ignore this, I will have to fix it

packages/react/src/refs.tsx
```diff
@@ -192,8 +200,26 @@ export function Template({
...
   const rootDataState = resolveRef(dataValues, path)
   return (
-    <BaseTemplate components={components} path={path} dataState={rootDataState} onEvent={onEvent} />
+    <BaseTemplate
+      components={components}
+      path={path}
+      dataState={rootDataState}
+      onAction={onAction}
+      onEvent={async (event) => {
+        const res = await onEvent(event)
```
grabbou, line 212: This could always call `dataSource.onEvent` to make sure it always work, but also expose the "onEvent" prop in form of side-effect

grabbou:
```ts
onEvent?.(event)
const res = await dataSource.sendEvent(event)
```

packages/ws-server/src/data-source.ts
```diff
@@ -178,5 +153,5 @@ export function createWSServerDataSource() {
     return () => handlers.delete(handler)
   }
 
-  return { update, onActionEvent, subscribe, get, updateRoot, handleWSConnection }
```
grabbou, line 181: Should we bring `onEvent` back, for side-effect console logging and debugging purposes? Would get called on each message, but would have no impact on the actual return to the client

packages/react/src/response.ts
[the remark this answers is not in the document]
grabbou: Submitted a new variant

packages/ws-server/src/data-source.ts
grabbou: Decided - YES

grabbou:
> I'm really not sure how much of this stuff should be strictly tied to every data source, we should discuss this

It is possible to make it data source specific so that each can define its own shape and style for responses in case there's different needs. What you return from your handler must be supported by currently used data source (and that is up to you). In our case, we have `statusCode` and `ok` because we use it to resolve/reject a promise. If you choose a different strategy in your custom data source, you may or may not need one.

I would keep it on the todo list to split/keep separate once we add additional data source. Otherwise, it's very tricky to do in advance.

packages/ws-client/src/data-source.ts
```diff
@@ -162,17 +156,15 @@ export function createWSDataSource(wsUrl: string): WebSocketDataSource {
...
-            if (promises.has(event.dataState.key)) {
-              reject(new Error('Request timeout'))
-              promises.delete(event.dataState.key)
-            }
-          }, event.dataState.timeout || 10_000)
-        })
-      }
+      return new Promise((resolve, reject) => {
+        promises.set(event.dataState.key, resolve)
+        setTimeout(() => {
```
ericvicenti, line 161: do we need to clear this timeout?

packages/ws-server/src/data-source.ts
ericvicenti: yes I think a lower-level onEvent would be great! we can do it in a follow-up PR

packages/react/src/response.ts
```diff
@@ -0,0 +1,38 @@
+import { ActionEventDataState, JSONValue } from './template'
+
+export const res = (payload: JSONValue): ServerResponseDataState => {
```
ericvicenti, line 3: usually I would do `export function` here

ericvicenti: Also I think we should have a more descriptive name. maybe `createResponse`?

ericvicenti (approved):
Nice!

packages/ws-client/src/data-source.ts
grabbou: I don't think so - but I am not an expert. There's a check if event has been already resolved inside the timeout, so it will get called for each event and just noop. Maybe it could be a good practice to set the timeout on the map and clear it.

packages/react/src/response.ts
grabbou: Sounds good!

grabbou: const vs let tho

packages/ws-client/src/data-source.ts
grabbou: Decided on Discord it's not a big deal for now.

packages/react/src/response.ts
```diff
@@ -0,0 +1,38 @@
...
+  return {
+    $: 'response',
+    payload,
+    // what are default values in `fetch`?
+    statusCode: 200,
```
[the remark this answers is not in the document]
grabbou: it's probably too much for now, but somehow wanted to indicate sucesfull response 😆

PR #48, opened 2024-05-10
feat: allow multiple actions

grabbou:


packages/react/src/__tests__/template.test.tsx
```diff
@@ -278,10 +285,7 @@ it('should accept event handler as a prop', () => {
         component: 'View',
         props: {
           ['data-testid']: 'button',
-          onClick: {
-            $: 'event',
-            action: 'navigate',
-          },
+          onClick: action('navigate'),
```
ericvicenti, line 288: no route argument? maybe we should allow the spread of arguments so you can call `action('navigate', 'routeName')` which will be converted to an array

ericvicenti (approved):

packages/react/src/__tests__/template.test.tsx
grabbou: Ah, yeah, this is just some random stuff for testing, I should probably not use same names as in production app to avoid confusion.

Let's do spread tho! It's quite handy.

PR #49, opened 2024-05-10
feat: allow events to have multiple side-effect actions

grabbou:


packages/react/src/refs.tsx
```diff
@@ -205,6 +205,9 @@ export function Template({
         onAction?.(event.dataState)
         return
       }
+      for (const action of event.dataState.actions) {
```
ericvicenti, line 208: is this safe when actions is undefined?

grabbou: it is always defined assuming it's a valid HandlerEventDataState. Such can only be created with `handler()` function that ensures this shape. We still have to implement runtime checking tho.

ericvicenti: ok gotcha. might want to add a check here just in case. Lots of people will hopefully be making server-side implementations in different languages! 😁

grabbou: True! However, I am not sure here is the best place to check for it. We should probably make `isEventDataState` function inside `<Template />` (type assertion) return false if it's missing some properties / not adhering to TypeScript type definition.

If we added a check here for `actions`, it would add a bit more safety but:
- TypeScript would report it as necessary because `actions` is always defined per type
- would not ensure other properties are correctly set, which is a bit inconsistent 

So, I would say we need to ensure that this is always correct shape.

grabbou: I opened up an issue https://github.com/final-ui/final/issues/50 to tackle this in a follow-up in a bit more comprehensive way. Let me know if you think this is good enough!

ericvicenti (approved):

PR #51, opened 2024-05-10
feat: add utility assertion for uniform arrays

grabbou:
Alternative to Array.prototype.every that throws if items are not all of the same type. That way, we can safely check for an array of "actions". Otherwise, current code would lead to [action(), "whatever"] to pass type checking, but lead to unexpected behavior during runtime.

After this PR, it will throw.


ericvicenti (approved):
LGTM

PR #52, opened 2024-05-11
feat(core): support multiple action events in a more elegant way

grabbou:
To define single action:

```ts
onPress={event(action('my-action'))}
```
or
```
{
  '$': 'event',
  actions: [
    { '$': 'action', name: 'my-action' }
  ]
}
```

To define multiple actions:
```ts
onPress={event([action('my-action'), action('my-other-action')])}
```
or
```
{
  '$': 'event',
  actions: [
    { '$': 'action', name: 'my-action' },
    { '$': 'action', name: 'my-other-action' }
  ]
}
```

To define multiple actions alongside an event handler:
```ts
onPress={event(() => { console.log('this runs on the backend') }, action('my-action'))}
```
or
```
{
    '$': 'event',
    key: 'f12ef653-5a7f-49a2-84e2-9953ccaa9ca4',
    handler: [Function (anonymous)],
    actions: [
      { '$': 'action', name: 'my-action' }
    ]
  }
```

Changes under the hood:
- We no longer allow "ActionDataState" to be passed as a prop, so you can't do "onPress={action()}". This simplifies the code and path handling this case together with assertion (recently introduced) was removed
- Once JSX lands, each function will be automatically wrapped with a `handler`, so the fact it supports both ActionDataState and function as an argument is future proof
- Introduced new type "$ action"

TemplateEvent can have dataState of either HandlerEventDataState or ActionEventDataState. ActionEventDataState is a data state that has property "actions" that is array of "ActionDataState" (with type "$"). HandlerEventDataState is a data state that has "key" and "handler". It can also have actions.


example/demo/src/inventory/ui.ts
```diff
@@ -38,7 +38,7 @@ export function getHomeScreen(): ServerDataState {
             component: 'Button',
             props: {
               unstyled: true,
-              onPress: action(['navigate', `inventory:${item.key}:details`]),
+              onPress: handler(action(['navigate', `inventory:${item.key}:details`])),
```
ericvicenti, line 41: why do we need `handler(action(` instead of just `action(`?

grabbou: `action` is just an optional shortcut for writing
```
{ $: 'action', name: '' }
```

Pretty much as per original post. Making action to be always part of handler let us remove some code on the client. We don't have to do special treatment for actions, but can treat them just like any other event. And I also didn't consider it too much of a trade-off given that we were going to abstract `handler` anyway.

Another reason is pure code clean-up
> TemplateEvent can have dataState of either HandlerEventDataState or ActionEventDataState. 

I reorganised data structures a bit in this PR, by introducing `$: action`. 

```ts
export type ActionEventDataState = {
  $: 'event'
  actions: ActionDataState[]
}
```

We now have `action event` that is an event with just actions. This is the result of calling `handler(action())` or `handler([action()])` and allows us to use same code for handling handler events as well as action events on the client.

Also, that is reused with:
```ts
export type HandlerEventDataState = {
  $: 'event'
  key: string
  actions?: ActionDataState[]
  timeout?: number
}
```
that can now also have actions associated with it (`handler(() => {}, action())`.

grabbou: > why do we need `handler(action(` instead of just `action(`?

We could add it, no problem. Behind the scenes, I would basically have to do this:
```ts
if (isActionDataState(propValue)) {
      return async (payload: any) => {
        return onTemplateEvent?.({
          target: {
            key: parentNode.key,
            component: parentNode.component,
            propKey,
            path,
          },
          dataState: {
              $: 'event',
              actions: [propValue]
          },
          payload,
        })
      }
    }
```
and same for array. As you can see, very similar to `handler`, it's just the:
```ts
dataState: {
    $: 'event',
    actions: [propValue]
},
```
that is different.

grabbou: Resolving this conversation for now in favour of updated PR description.

ericvicenti (approved):
the helpers are very clean now!

PR #54, opened 2024-05-16
feat: add experimental http data source 

grabbou:
Adds new example and ability to deploy it
Uses SQL to load data
Implements lazy initialisation of a page


packages/playground/src/data-sources.tsx
```diff
@@ -1,13 +1,21 @@
+import { createHTTPDataSource } from '@final-ui/http-client'
 import { createWSDataSource } from '@final-ui/ws-client'
 
 const dataSources = new Map<string, ReturnType<typeof createWSDataSource>>()
 
+function createDataSource(url: string) {
+  if (url.match(/^ws:\/\//)) {
```
ericvicenti, line 7: this should probably use a real url parser and check for the protocol/scheme.

This is also a bug because it does not check for wss

ericvicenti (approved):

PR #56, opened 2024-05-16
feat: rework events to not rely on a key, but a path, so we can use lazy initialisers

grabbou:
We still generate random key on the server, so we can use it to keep track of a promise. Didn't want to require client-side code to pull in crypto or something.


packages/react/src/__tests__/events.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+  `)
+
+  // prop is a single child
+  fireEvent.click(component.getByTestId('button-prop'))
+  expect(onTemplateEvent.mock.lastCall[0].target.path).toMatchInlineSnapshot(`
+    Array [
+      "",
+      "children",
+      "children",
+      0,
```
ericvicenti, line 97: we should support and encourage use of keys here, so if the event happens and the order has changed on the server, the correct event gets called

packages/react/src/template.tsx
```diff
@@ -192,22 +191,22 @@ export function BaseTemplate({
...
-          return [key, renderProp(key, value, parentNode, path)]
+          return [key, renderProp(key, value, parentNode, [...path, key])]
         })
       )
     }
     return propValue
   }
 
-  return <>{render(dataState, path)}</>
+  return <>{render(dataState, Array.isArray(path) ? path : [path])}</>
```
ericvicenti, line 209: not necessary

ericvicenti (approved):
great improvements!

PR #58, opened 2024-05-17
feat: implement events for http data source (initial take) 

grabbou:
Still long way to go, but it should be good enough.


ericvicenti (approved):
nice, happy to see events support in http!

grabbou:
Write tests

grabbou:
@ericvicenti removed "cache" from data source that I added in this PR. It leads to some issues. Component enters infinite loop -> mount - fetch - set cache - unmount - remove cache (last listener). On next mount, it fetches again because cache is empty.

I am removing cache till we have better mechanism to handle this + revalidation.

grabbou:
Also, we should figure what is the best "default" value for path. For HTTP data source, it makes most sense to be "/". This is because, when you call a URL: "http://server.com:3015" the pathname on the server is implicitly "/", even though it was "" on the client. To force it to be "" on the client, we would have to pass it as a query param. This leads to issues between what's defined on the server (dataSource.update("", value)) vs. what is received (dataSource.get("/"), value = "undefined").

For now, to use experimental HTTP, you must add the following `http://localhost:3015` and `/` as path. 

**Apart from that, events and fetching work.**

grabbou:
FYI unified paths between ws and http example - both are "" and "items" etc., without leading "/". This makes it much easier and actually removes a few bugs.

Will add some more unit tests later, but it's good to go.

PR #61, opened 2024-05-22
fix: collision when firing events concurrently

### LINKED ISSUE #40: bug: do not generate uuid of event handler on the server, but on the client
We should generate uuid each time there's interaction. Not once, on the server. Right now, there will be an error if you fire too many interactions at once, b/c multiple events will point towards single event handler UUID.

If button is tapped 5 times, each of those events should have unique identifier for the callback on the server.

grabbou:
Fixes #40

Introduces new type of DataState - "ActionsDataState"
Generates key dynamically for each event, when putting promise on a map


ericvicenti (approved):
Looks great

PR #62, opened 2024-05-27
feat: local state 

grabbou:
This PR adds ability to express local state and updates to it as a part of JSON schema. The following scenarios are unit-tested and currently supported:
- [x] render initial state
- [x] set state to value from payload (onPress={setStateAction()})
- [x] set state with custom value (onPress={setStateAction(state, 'hello world')})
- [x] set state with a modifier (onPress={setStateAction(state, toggle)})
- [x] set state with a modifier that references previous value (tested with multiple events to ensure they're batched) (2x onPress={setStateAction(state, increment(2)})
- [x] set state in return from the server (onPress={() => return setStateAction(state, toggle)})

----

The structure is as follows:
```ts
const isDisabled = state(false) 
```

You use `state` helper to create StateDataState:
```ts
{ $: 'state', key: '<generated>', initialValue: false }
```

You can pass that value as a prop to anything where you want it to be substituted with actual state value on the client, e.g.:
```jsx
<Button disabled={isDisabled} />
```

In order to express a state update, you can use an action:
```jsx
<Button onPress={setStateAction(isDisabled, false)} />
``` 

When you don't specify second argument to `setStateAction`, it will be assigned the value that was passed to the handler. This is useful for inputs:
```jsx
<TextInput onChangeText={setStateAction(userName)} />
```

You can also pass a `StateModifier` as a second argument to express advanced operations that depend on state value, such as `toggle` (to disable/enable, depending on current value, or increment).
```jsx
<Button onPress={setStateAction(isDisabled, toggle)} />
```
```jsx
<Button onPress={setStateAction(isDisabled, increment(2))} />
```

`toggle`, `increment` and `setStateAction` are helper functions to generate JSON-compliant definitions:
```ts
type UpdateStateAction = {
  $: 'action'
  name: ['state-update', stateKey, newStateValueOrModifier]
}
 
type StateModifier = {
  $: 'state-modifier'
  value: 'payload' | 'toggle' | ['increment', number]
}
```

Since state update is an action, you can also return it from the server-side handler, to e.g. set validation error in response from remote server function:
```jsx
<Form
   onSubmit={() => {
      // perform your validation
      return setStateAction(errors, { name: "This field is required" })
   }}
/>
```

Important: The last example with response needs some improvements, I have described them here: https://github.com/final-ui/final/issues/63

and here: https://github.com/final-ui/final/issues/64 

These two, but most importantly #64 are required to support fully server-side defined kit components.


packages/react/src/template.tsx
```diff
@@ -174,6 +206,7 @@ export function BaseTemplate({
     propKey: string,
     propValue: DataState,
     parentNode: ComponentDataState | ReferencedComponentDataState,
+    localState: LocalState,
```
grabbou, line 209: We need to pass this down from component that calls `renderProp()`, unless we have `use()` hook from canary that allows to access context from helper functions / loops.

packages/react/src/template.tsx
```diff
@@ -117,36 +130,49 @@ export function BaseTemplate({
...
-    if (!componentDefinition) {
-      throw new RenderError(`Unknown component: ${stateNode.component}`)
-    }
-
-    // tbd: validate `components` prop once with `zod` instead of later render stage
-    const Component = componentDefinition.component
-    if (!Component) {
-      throw new RenderError(`Invalid component: ${stateNode.component}`)
-    }
+  const RenderComponent = useCallback(
```
grabbou, line 133: This was causing TextInput to loose focus, because this function was re-created each render, resulting in a whole new component being put on screen (unmount input, mount a new one). Likely the reason why some BottomSheet issues were happening!

packages/react/src/__tests__/state.test.tsx
```diff
@@ -0,0 +1,273 @@
...
+            component: 'View',
+            children: value,
+          },
+          {
+            $: 'component',
+            key: 'button',
+            component: 'View',
+            props: {
+              ['data-testid']: 'button',
+              onClick: event(setStateAction(value)),
```
ericvicenti, line 54: I'm still not convinced the user should need to wrap with `event()` here

packages/react/src/__tests__/state.test.tsx
```diff
@@ -0,0 +1,273 @@
...
+  const value = state('foo')
+  const dataSource: DataSource = {
+    get: () => ({
+      subscribe: () => jest.fn(),
+      get() {
+        return [
+          {
+            $: 'component',
+            key: 'header',
+            component: 'View',
```
ericvicenti, line 92: this should probably be a `<Text>`

packages/react/src/__tests__/state.test.tsx
```diff
@@ -0,0 +1,273 @@
...
+            children: counter,
+          },
+          {
+            $: 'component',
+            key: 'button',
+            component: 'View',
+            props: {
+              ['data-testid']: 'button',
+              onClick: event(setStateAction(counter, increment(2))),
```
ericvicenti, line 197: nice!

packages/react/src/__tests__/state.test.tsx
```diff
@@ -0,0 +1,273 @@
...
+          component: 'View',
+          props: {
+            ['data-testid']: 'button',
+            onClick: event(jest.fn()),
+          },
+          children: value,
+        }
+      },
+    }),
+    sendEvent: jest.fn().mockReturnValue(response(null).action(setStateAction(value, 'bar'))),
```
ericvicenti, line 249: hmm I'll think about the `response()` API.. in any case it should maybe have a more specific label like `eventResponse` so that it is more clear what the purpose is

packages/react/src/state.tsx
```diff
@@ -0,0 +1,98 @@
...
+export function state<T extends JSONValue>(initialValue: T): StateDataState<T> {
+  const key = (Date.now() * Math.random()).toString(16)
+  return { $: 'state', key, initialValue }
+}
+
+export function setStateAction<T>(
+  state: StateDataState<T>,
+  value: T | StateModifier = eventPayload
+): UpdateStateAction<T> {
+  return action(['state-update', state, value])
```
ericvicenti, line 30: I guess this will be the string convention of all our built in features, `kebab-case`? I think we used camelCase in other cases like goBack? Just want to make sure we have consistency

packages/react/src/state.tsx
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export const toggle: StateModifier = {
+  $: 'state-modifier',
+  value: 'toggle',
+}
+
+export const increment = (value: number): StateModifier => ({
+  $: 'state-modifier',
+  value: ['increment', value],
```
ericvicenti, line 95: why use a tuple here instead of a seperate value?

packages/react/src/state.tsx
```diff
@@ -0,0 +1,98 @@
+import { createContext, useReducer } from 'react'
+
+import { action } from './events'
+import { ActionDataState, JSONValue, StateDataState } from './template'
+
+export type LocalState = Record<string, JSONValue>
+
+type StateUpdate<T> = T | StateModifier
+type UpdateStateAction<T> = ActionDataState<['state-update', StateDataState<T>, StateUpdate<T>]>
+type StateModifier = {
```
ericvicenti, line 10: usually I would use a union type here `PayloadStateModifier | ToggleStateModifier` etc

packages/react/src/state.tsx
```diff
@@ -0,0 +1,98 @@
...
+    }
+    case 'toggle': {
+      return !state
+    }
+    default: {
+      if (Array.isArray(stateUpdate.value)) {
+        const [type, value] = stateUpdate.value
+        switch (type) {
+          case 'increment': {
+            // @ts-ignore what to do if `state` is not a number? same in other cases
```
ericvicenti, line 73: yeah ideally we could use something like zod in any case where we have API data that we need to validate

packages/react/src/template.tsx
ericvicenti: oh!! I knew it would be something simple! thank you for finding this, it has caused a lot of pain

packages/react/src/template.tsx
```diff
@@ -117,36 +130,49 @@ export function BaseTemplate({
...
+            propKey,
+            renderProp(propKey, propValue, stateNode, localState, [...path, 'props', propKey]),
+          ]
+        })
+      )
+      if (typeof componentDefinition.validator === 'function') {
+        try {
+          componentProps = componentDefinition.validator(componentProps)
+        } catch (e) {
+          throw new RenderError(
```
ericvicenti, line 158: unrelated to this change but we probably want to pass in the validation error as a first-class value so that the suspense boundary can have nice logic for addressing the error

packages/react/src/template.tsx
ericvicenti: use will be very handy!

also, have we considered using a different context for each state, so that only the relevant components update when a state changes?

packages/react/src/template.tsx
```diff
@@ -99,6 +109,9 @@ export function isActionsDataState(obj: DataState): obj is ActionsDataState {
 export function isActionDataState(obj: any): obj is ActionDataState {
   return obj !== null && typeof obj === 'object' && '$' in obj && obj.$ === 'action'
 }
+function isStateDataState(obj: any): obj is StateDataState {
+  return obj !== null && typeof obj === 'object' && '$' in obj && obj.$ === 'state'
```
ericvicenti, line 113: btw I don't think it is necessary to check for `'$' in obj`

ericvicenti (approved):
LFG!!

packages/react/src/__tests__/state.test.tsx
grabbou: When you use JSX syntax, you don't have to. Haven't configured this in the unit test, so I am doing it explicitly.

packages/react/src/__tests__/state.test.tsx
grabbou: Yeah, I am sticking to a single "View" component created for the sake of unit tests, we could create Text that is "p" (or "span") too, to improve readability, although it is not necessary.

packages/react/src/__tests__/state.test.tsx
grabbou: #63 

Yeah. The API should change, because it's not as good now as the library is growing. I have this open issue and plan to tackle is as a next thing, after rewriting all kit components to be server-only.

packages/react/src/state.tsx
grabbou: Could be! I somewhat defaulted to kebab case and forgot Rise was camelCase. Technically, Rise is a separate project so it can have its own action names, but I think we should have a single way and I am fine with kebab case (refactor other actions in separate PR?)

packages/react/src/state.tsx
grabbou: So, alternative would be:
```
{ $: 'state-modifier', type: 'increment', value: JSONValue }
```
and for others, the `value` would not be present. I think that's better design and will simplify switch/case. Will update!

packages/react/src/state.tsx
grabbou: Sounds good. Let's change to union type.

packages/react/src/state.tsx
grabbou: In this case, however, should we just ignore the TS error and perform addition? In case it's not a number, it will be NaN, but I think that is expected when you do this in pure JS either.

packages/react/src/template.tsx
grabbou: TypeScript complains otherwise for `obj.$` :(

packages/react/src/template.tsx
grabbou: will turn this into issue - haven't thought much about error handling, so don't know how to answer this yet, but it's important for sure.

grabbou:
https://github.com/final-ui/final/pull/62#discussion_r1623169797

Follow-up in #66

packages/react/src/state.tsx
grabbou: Ended up throwing an error if it's not a number.

packages/react/src/template.tsx
ericvicenti: Yeah definitely important pre v1 release, but maybe we don't have to worry for the beta (release for the first conf)

packages/react/src/__tests__/state.test.tsx
ericvicenti: yeah, no big deal because that isn't what we are testing

packages/react/src/__tests__/state.test.tsx
ericvicenti: hmm. but I wonder about the actual JSON definition, it should be fine to provide an action directly as a handler, right?

packages/react/src/__tests__/state.test.tsx
grabbou: thanks! it's also type-safe, so you can't pass `toggle` to a string state or `increment`. Only to what is compatible. If you ignore it, you will receive some strange (most likely) outcomes!

packages/react/src/state.tsx
grabbou: I am going to default to kebab-case and convert the rest in a follow-up

packages/react/src/__tests__/state.test.tsx
grabbou: So actual JSON definition allows for two things:
```
{$: 'event', handler: () => {}, actions: []}
```
^ an event that has function to execute on the server
and
```
{$: 'actions', actions: []}
```
like event, but is local only.

When you call `event(something)`, that helper chooses which thing to return based on arguments. If you do `event(action)` (no handler), it returns `{$: 'actions', actions: []}`. If you do `event(fn)`, it does the former. 

We chose not to support passing a single action or an array of actions, to avoid looping over an entire array to make sure all elements are of the same type. Instead, users have to wrap them with `{$: 'actions', actions: []}`.

grabbou: We could allow passing a single action, or an array of actions (requires checking each element of an array is of the same type), now that I think about it.

We would then, on the client, essentially wrap them with `{$: 'actions', actions: []}`, since that's already what our code expects.

ericvicenti: hmm, well a single action doesn't need the array or `$: 'actions'`

grabbou: Yeah we can always add support for passing single action. But it's only going to be useful for those not using JSX and using raw JSON.

PR #70, opened 2024-06-01
refactor: move from Cloudflare to Fastify and support more servers 

grabbou:
Follow-up to https://github.com/final-ui/final/commit/9ba7f8f3395b779c029f04cbb38e56c61e63aa21 that does the same, but for web sockets


ericvicenti (approved):
Thanks for the clean up

I plan on rewriting the servers soon but it makes me happy to have a clean environment again. One day we can do a `cf-demo` or something

PR #71, opened 2024-06-01
feat: refactor `event` API to support timeout and allow action to be passed in directly

### LINKED ISSUE #69: feat: support single action as a part of JSON spec
Relevant discussion here: https://github.com/final-ui/final/pull/62#discussion_r1623223980

CC: @ericvicenti

grabbou:
Fixes #69

### DIFF
--- a/packages/react/src/__tests__/template.test.tsx
+++ b/packages/react/src/__tests__/template.test.tsx
@@ -293,15 +293,12 @@ it('should accept event handler as a prop', () => {
   const firedEvent = onEvent.mock.lastCall[0] as TemplateEvent
   expect(firedEvent).toMatchInlineSnapshot(`
     Object {
-      "dataState": Object {
-        "$": "actions",
-        "actions": Array [
-          Object {
-            "$": "action",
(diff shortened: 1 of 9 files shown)


packages/react/src/__tests__/template.test.tsx
```diff
@@ -378,7 +375,7 @@ it('should pass return type from onTemplateEvent back to component', async () =>
         key: 'button',
         component: 'Profile',
         props: {
-          onClick: event(action('go-back')),
+          onClick: action('go-back'),
```
ericvicenti, line 378: nice much cleaner. This will also support JSX right? So I can do `onClick={action('go-back')}` ?

ericvicenti (approved):
Ok definitely seems like an improvement!

packages/react/src/__tests__/template.test.tsx
grabbou: yes! it is natively supported, so you can pass this value as a prop with/without JSX and it will work just fine (it is not wrapped in event anymore behind the scenes).

PR #73, opened 2024-06-02
feat: refactor ServerDataState and DataState types as well as allow populating event values with stream 

grabbou:
Except for one broken unit test (outdated values), everything works lol

### DIFF
--- a/packages/react/src/__tests__/events.test.tsx
+++ b/packages/react/src/__tests__/events.test.tsx
@@ -1,7 +1,6 @@
 import { fireEvent, render } from '@testing-library/react'
 import React, { PropsWithChildren, ReactNode } from 'react'
 
-import { event } from '../events'
 import { BaseTemplate } from '../template'
 
 const View = {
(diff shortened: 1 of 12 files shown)


packages/react/src/__tests__/events.test.tsx
```diff
@@ -51,13 +50,17 @@ it('should assign correct path to an event target', () => {
               key: 'container',
               props: {
                 ['data-testid']: 'button',
-                onClick: event(jest.fn()),
+                onClick: {
+                  $: 'event',
+                },
```
grabbou, line 55: We're defining client-side DataState in unit tests and here, `handler` is not present (functions are not part of JSON after stringifying them). So, we're not going to use the `event()` helper, but rather define some actual JSON.

packages/react/src/__tests__/state.test.tsx
```diff
@@ -271,3 +272,83 @@ it('should modify state after receiving response from the server', async () => {
...
+          },
+        ]
+      },
+    }),
+    sendEvent: onTemplateEvent,
+  }
+  const component = render(<Template components={BUILT_IN_COMPONENTS} dataSource={dataSource} />)
+  await act(async () => {
+    fireEvent.click(component.getByTestId('checkbox'))
+    fireEvent.click(component.getByTestId('button'))
```
grabbou, line 351: this test is broken - firing concurrent events in a single loop. Need to fix it.

packages/react/src/jsx.tsx
```diff
@@ -22,13 +23,15 @@ type Props = Record<string, JSONValue | AllowedDataStates> & {
...
+
 export const jsxs = jsx
 
 export function jsx(
-  componentFactory: (props: any) => ReactElement<Props> | ComponentDataState,
+  componentFactory: (props: any) => UI,
   { children, ...passedProps }: Props,
   key?: string
-): ComponentDataState {
+): ComponentDataState<ServerDataState> {
```
grabbou, line 34: Making type definitions a bit tighter now that ComponentDataState is a generic!

PR #74, opened 2024-06-02
feat: js server atomic model architecture

ericvicenti:


packages/server/src/ws-connection.ts
```diff
@@ -0,0 +1,184 @@
...
+  })
+
+  function handleSubKey(key: string) {
+    // console.log(`${clientId} sub to key ${key}`)
+    async function send() {
+      const sender = clientSenders.get(clientId)
+      if (!sender) {
+        return
+      }
+      const model = getModel(key) as ValueModel<unknown> | undefined
```
grabbou, line 88: Why casting here is necessary?

packages/server/src/ws-connection.ts
```diff
@@ -0,0 +1,184 @@
...
+      const val = typeof model === 'function' ? model() : model?.get()
```
grabbou, line 89: nit: could we use getModelState here?

packages/server/src/ws-connection.ts
```diff
@@ -0,0 +1,184 @@
...
+      })
+    }
+    const handlers =
+      clientSubscribers.get(key) ||
+      (() => {
+        const handlers = new Map<string, () => void>()
+        clientSubscribers.set(key, handlers)
+        const model = getModel(key)
+        if (model && typeof model === 'object') {
+          model.subscribe(() => {
```
grabbou, line 104: Should we clean-up this subscription as well?

packages/server/src/ws-connection.ts
```diff
@@ -0,0 +1,184 @@
...
+  keys: z.array(z.string()),
+})
+const serverEventMessageSchema = z.object({
+  $: z.literal('evt'),
+  event: z.object({
+    target: z.object({
+      key: z.string().optional(),
+      component: z.string(),
+      propKey: z.string(),
+      path: z.array(z.string().or(z.number())),
```
grabbou, line 28: nit: 
```
z.tuple([z.string()]).rest(z.string().or(z.number()))
```
will give
```
type Path = [string, ...(string | number)[]] 
```
which will let you remove some `if` conditions, typically:
```
if (!storeName) throw new Error('Missing store name in event path')
if (typeof storeName !== 'string') throw new Error('Store name must be a string')
```

packages/server/src/ws-connection.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  async function handleEvt({ key, event }: z.infer<typeof serverEventMessageSchema>) {
+    // console.log('handleEvt', key, event)
+    const { target, payload } = event
+    const { path } = target
+    try {
+      const [storeName, ...lookupPath] = path
+      if (!storeName) throw new Error('Missing store name in event path')
+      if (typeof storeName !== 'string') throw new Error('Store name must be a string')
+      const model = findModel(models, storeName.split('/'))
```
grabbou, line 136: nit: should probably call `getModel` from context here

packages/server/src/ws-connection.ts
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  // console.log('connected client', clientId)
+  ws.on('message', (messageData) => {
+    let messageUnvalidated, message
+    try {
+      messageUnvalidated = JSON.parse(messageData.toString())
+      // console.log(messageUnvalidated)
+    } catch (e) {
+      throw new Error('Failed to parse JSON from client ' + clientId)
```
grabbou, line 170: nit: might be worth writing e.message or something to better understand what caused the failure.

To be honest, I think one Error: "Failed to process message" could also work for both JSON.parse and Zod.parse, since error message would be the same otherwise.

packages/ws-client/src/data-source.ts
```diff
@@ -122,19 +122,21 @@ export function createWSDataSource(wsUrl: string): WebSocketDataSource {
...
-        if (shouldSubscribeRemotely) {
+        if (shouldSubscribeRemotely && key) {
+          console.log('client sub', key)
           send({
             $: 'sub',
             keys: [key],
           })
         }
         return () => {
+          console.log('client unsub', key)
```
grabbou, line 133: nit: we could probably remove some of these logs!

packages/server/src/ws-server.ts
```diff
@@ -0,0 +1,20 @@
+import { WebSocketServer } from 'ws'
+
+import { AnyModels } from './types'
+import { connectWebSocket, createWSServerContext } from './ws-connection'
+
+export function createWSServer(models: AnyModels, port: number) {
+  const wss = new WebSocketServer({ port })
+
```
grabbou, line 8: nit: looking at the PR overall I like your concise style with no white spaces, so I would be down getting rid of them in other files like this one too! not a big deal tho!

packages/ws-client/src/data-source.ts
```diff
@@ -65,7 +65,7 @@ export function createWSDataSource(wsUrl: string): WebSocketDataSource {
 
   rws.addEventListener('open', () => {
     const keys = [...subscriptions.entries()]
-      .filter(([, handlers]) => handlers.size > 0)
+      .filter(([key, handlers]) => !!key && handlers.size > 0)
```
grabbou, line 68: Can key in a map be undefined? Please explain!

packages/server/src/model-view.ts
```diff
@@ -0,0 +1,60 @@
...
+          isInvalid = true
+        })
+      )
+      if (model.type === 'query') {
+        return model.get()
+      } else if (model.type === 'state') {
+        return model.get()
+      } else if (model.type === 'view') {
+        return model.get()
+      }
```
grabbou, line 31: Maybe we could use getModelState helper here too? Sounds like it handles this case + it's .get() for each model anyway right now.

grabbou (approved):
Looks good! 🚀 Love the change - overall very smart and concise design! 🔥 

 some comments to better understand few areas & nits to either unify styling or to better understand rationale behind. Overall it's good to go!

grabbou:
Before hitting that "Squash and merge" button, let's make sure we have PR description and title right! 🚨🚨

packages/ws-client/src/data-source.ts
ericvicenti: I don't remember, probably to make ts happy

packages/server/src/model-view.ts
ericvicenti: yeah sure

example/demo/src/inventory/ui.tsx
```diff
@@ -72,13 +96,19 @@ function HomeScreen() {
...
+  item: Item
+  onUpdateInventory: StateSetter<Item[]>
+}) {
   return (
     <YStack flex={1} backgroundColor={'$background'} gap="$3">
       <Image
         key="photo"
         source={{
-          uri: ref([`inventory-items`, item.key, 'photo']),
+          uri: ref([`inventoryItems`, item.key, 'photo']),
```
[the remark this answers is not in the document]
ericvicenti: yeah either way is fine, I just want to make sure we have some active example code that uses refs

packages/server/src/__tests__/model-lookup.test.ts
```diff
@@ -0,0 +1,18 @@
...
+      return state(key)[0]
+    })
+    expect(l.get('a')?.get()).toBe('a')
+  })
+  test('lookup will cache keys', () => {
+    const l = lookup((key) => {
+      return state(key)[0]
+    })
+    const a = l.get('a')
+    expect(a).toBe(l.get('a'))
```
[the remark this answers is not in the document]
ericvicenti: this checks that the exact same model gets returned from lookup. it does check caching

packages/server/src/model-utils.ts
```diff
@@ -0,0 +1,46 @@
+import { AnyModels, ValueModel } from './types'
+
+export function findModel<V>(
+  models: AnyModels,
+  path: string[]
+): undefined | ValueModel<V> | (() => V) {
+  let walkModel: AnyModels | undefined = models
+  path.forEach((term) => {
```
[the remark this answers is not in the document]
grabbou: for of

packages/server/src/model-utils.ts
```diff
@@ -0,0 +1,46 @@
...
+      // cannot look up an item in a query. maybe drill into the value here?
+      walkModel = undefined
+    } else if (walkModel.type === 'view') {
+      // cannot look up an item in a view. maybe drill into the value here?
+      walkModel = undefined
+    } else {
+      walkModel = walkModel[term]
+    }
+  })
+  while (walkModel && typeof walkModel !== 'function' && walkModel?.type === 'lookup') {
```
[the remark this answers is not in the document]
grabbou: + would be cool to see a very simple test demonstrating when this gets called!

packages/server/src/model-utils.ts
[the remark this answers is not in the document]
ericvicenti: will add a test here

packages/server/src/model-utils.ts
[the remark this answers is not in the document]
grabbou: Let’s also add unit test here (very simple) to remember this case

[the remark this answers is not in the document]
ericvicenti: will write a test

packages/server/src/ws-connection.ts
ericvicenti: lets skip this for now and fix it when we separate the model path from value path

packages/server/src/ws-connection.ts
ericvicenti: apparently not necessary!

packages/server/src/ws-connection.ts
ericvicenti: yes that should be possible! will try to use it here

packages/server/src/ws-connection.ts
ericvicenti: true true! will try to fix this

packages/ws-client/src/data-source.ts
ericvicenti: oh I remember, key was null here for some reason and the server was breaking. we should find the root cause

PR #75, opened 2024-06-02
feat: allow referencing local state values within the function handler

grabbou:
This PR allows referring local state in the event handler, as follows:
```tsx
function Component() {
  const userName = state('')

  return (
    <>
      <TextInput onTextChange={setStateAction(userName)} />
      <Button
        onPress={event(
          (state) => {
            console.log(state.userName) // Mike
          },
          { args: { userName } }
        )}
      />
    </>
  )
}
```

### DIFF
--- a/packages/react/src/__tests__/events.test.tsx
+++ b/packages/react/src/__tests__/events.test.tsx
@@ -1,7 +1,6 @@
 import { fireEvent, render } from '@testing-library/react'
 import React, { PropsWithChildren, ReactNode } from 'react'
 
-import { event } from '../events'
 import { BaseTemplate } from '../template'
 
 const View = {
@@ -51,13 +50,17 @@ it('should assign correct path to an event target', () => {
               key: 'container',
               props: {
                 ['data-testid']: 'button',
-                onClick: event(jest.fn()),
+                onClick: {
+                  $: 'event',
+                },
                 header: {
                   $: 'component',
                   component: 'View',
                   props: {
                     ['data-testid']: 'button-prop',
-                    onClick: event(jest.fn()),
+                    onClick: {
+                      $: 'event',
+                    },
                   },
                 },
                 footer: [
@@ -67,7 +70,9 @@ it('should assign correct path to an event target', () => {
                     component: 'View',
                     props: {
                       ['data-testid']: 'button-prop-array',
-                      onClick: event(jest.fn()),
+                      onClick: {
+                        $: 'event',
+                      },
                     },
                   },
                   {
@@ -75,7 +80,9 @@ it('should assign correct path to an event target', () => {
                     component: 'View',
                     props: {
                       ['data-testid']: 'button-prop-array-idx',
-                      onClick: event(jest.fn()),
+                      onClick: {
+                        $: 'event',
+                      },
                     },
                   },
                 ],
--- a/packages/react/src/__tests__/state.test.tsx
+++ b/packages/react/src/__tests__/state.test.tsx
@@ -1,7 +1,6 @@
 import { act, fireEvent, render } from '@testing-library/react'
 import React from 'react'
 
-import { event } from '../events'
 import { DataSource, Template } from '../refs'
 import { response } from '../response'
 import { increment, setStateAction, state, toggle } from '../state'
@@ -240,7 +239,9 @@ it('should modify state after receiving response from the server', async () => {
           component: 'View',
           props: {
             ['data-testid']: 'button',
-            onClick: event(jest.fn()),
+            onClick: {
+              $: 'event',
+            },
           },
           children: value,
         }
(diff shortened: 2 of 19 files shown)


packages/react/src/__tests__/events.test.tsx
```diff
@@ -51,13 +50,17 @@ it('should assign correct path to an event target', () => {
               key: 'container',
               props: {
                 ['data-testid']: 'button',
-                onClick: event(jest.fn()),
+                onClick: {
+                  $: 'event',
```
grabbou, line 54: I made types more strict. `event` returns `ServerEventDataState` (it has property "handler" that can be a function) and that property is obviously gone on the client after serialisation. So, we can't use that utility here, because it would make TypeScript complain.

This is client-side JSON definition passed directly to React component, so we have to make sure it does not have any server-specific properties (just like in real life).

packages/react/src/__tests__/state.test.tsx
```diff
@@ -271,3 +272,83 @@ it('should modify state after receiving response from the server', async () => {
...
+          },
+        ]
+      },
+    }),
+    sendEvent: onTemplateEvent,
+  }
+  const component = render(<Template components={BUILT_IN_COMPONENTS} dataSource={dataSource} />)
+  await act(async () => {
+    fireEvent.click(component.getByTestId('checkbox'))
+    fireEvent.click(component.getByTestId('button'))
```
grabbou, line 351: This is broken test case - need to fix state updates when concurrent event happen in a single tick.

packages/react/src/events.ts
```diff
@@ -39,3 +45,14 @@ export function action<T = any>(name: T): ActionDataState<T> {
     name,
   }
 }
+
+export function withState<T extends Record<string, any>>(
+  state: { [K in keyof T]: StateDataState<T[K]> },
```
grabbou, line 50: this makes it so that arguments to your `fn` are exactly as the keys/values passed as `state`.

ericvicenti: nice. Should we do `Record<string, unknown>` so that TS will continue inferring?

packages/react/src/__tests__/events.test.tsx
ericvicenti: great, I know this stuff is tricky but it seems like a big step in the right direction. Definitely can't use the same state for client and server.

packages/react/src/events.ts
grabbou: Surprisingly, it does infer it 100% perfect even though there's `any`!

packages/react/src/__tests__/state.test.tsx
grabbou: This test case is fixed in #76

packages/react/src/events.ts
```diff
@@ -39,3 +40,14 @@ export function action<T = any>(name: T): ActionDataState<T> {
...
+  func: HandlerFunction<T>
+): ServerHandlerDataState {
+  return {
+    $: 'handler',
```
ericvicenti, line 49: hm, can you clarify the difference between $: "handler" and "event"?

packages/react/src/template.tsx
```diff
@@ -47,11 +47,31 @@ export type ActionDataState<T = any> = {
...
 }
+export type ResponseDataState = {
+  $: 'response'
+  payload: JSONValue
+  statusCode: number
+  ok: boolean
+  actions: ActionDataState[]
+}
+export type HandlerDataState = {
+  $: 'handler'
```
ericvicenti, line 58: oh so this is used for the server state only? and it gets converted to an event before sent to the client?

packages/react/src/events.ts
grabbou: handler is just a special type that can be passed to `event`. Event takes either function or this object. When this object is present, it means you want to populate args of the `func` with `state`. Perhaps we could think of a better name, e.g. `func-with-state` or similar?

packages/react/src/template.tsx
```diff
@@ -47,11 +47,31 @@ export type ActionDataState<T = any> = {
...
+  state: Record<string, StateDataState>
+}
+export type HandlerReturnType = ResponseDataState | JSONValue | void
+export type HandlerFunction<T = any> = (args: T) => Promise<HandlerReturnType> | HandlerReturnType
 export type EventDataState = {
   $: 'event'
+  handler?: HandlerDataState
```
ericvicenti, line 65: maybe we should not have handler on the event. instead we could have:

```
type ClientState = EventState | ...
type HandlerState = { $: 'handler', handler: fn... }
type ServerState = ClientState | HandlerState
```

Then the system will automatically convert from ServerState to ClientState by converting HandlerState into EventState

grabbou: Could you elaborate a bit more on your example? I am not sure I got it right!

To further explain, the idea here was to scope changes to `fn`, without affecting the rest of the logic. So, before this PR, `event` only accepted a function `event(() => {})`. 

After this PR, `event` also takes `HandlerDataState` that can be instantiated with `withState`, e.g. `event(withState({}, () => {}))`, which allows us to modify the behavior of the function called. 

You can still use `event` tho to specify local actions, timeout etc., everything that you'd normally do anyway.

packages/react/src/events.ts
ericvicenti: hmm, maybe I was confused because this function returns a "ServerHandlerDataState" when it is not really a "DataState"

I thought a DataState is the thing sent to the client, but this is not ever sent to the client, it is only used internally to the server

packages/react/src/template.tsx
grabbou: Removed this so ignore.

packages/react/src/template.tsx
grabbou: I have removed this

packages/react/src/events.ts
grabbou: Agreed! Removed that type as well, and in the future, we might need a better name to differentiate across types that may loose certain properties while being serialised.

packages/react/src/jsx.tsx
```diff
@@ -1,16 +1,18 @@
...
   DataState,
   isComponentDataState,
   JSONValue,
   ReferencedDataState,
+  ServerEventDataState,
   StateDataState,
 } from './template'
 
+// tbd: refactor and unify this
 type AllowedDataStates =
```
grabbou, line 16: This should not be needed, but should match `ComponentDataState['props']`. Right now, ComponentDataState is more relaxed, and this is stricter, so we need to unify. Not a big deal for now, but needs to be done.

ericvicenti (approved):
🚀

PR #76, opened 2024-06-02
feat: move state to streams and populate payload with state values 

grabbou:
Now, only components that `useStream` will re-render. 

Unfortunately, there's one challenge that I highlighted in relevant line. Opening a sub-PR against other branch to make it easier to focus on this challenge alone.

### DIFF
--- a/packages/react/src/state.tsx
+++ b/packages/react/src/state.tsx
@@ -1,7 +1,7 @@
 import { createContext, useRef } from 'react'
 
 import { action } from './events'
-import { createWritableStream, Stream } from './streams'
+import { createWritableStream, Stream, WritableStream } from './streams'
 import { ActionDataState, JSONValue, StateDataState } from './template'
 
 type StateUpdate<T> = T | StateModifier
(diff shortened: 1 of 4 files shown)


packages/react/src/template.tsx
```diff
@@ -269,7 +263,8 @@ export function BaseTemplate({
       )
     }
     if (isStateDataState(propValue)) {
-      return render(localState[propValue.key] || propValue.initialValue, path)
+      // tbd: how to make this work?
+      return render(localState.getStream(propValue).get(), path)
```
grabbou, line 267: This worked previously, because all components re-run on `localState` change. Now, only ones that access state with `useStream` will re-render. 

As a result, props don't refresh on state change. 

There's few ways this can be done, and I think one of them are:
- pass in special object (instead of `localState`) that returns value from `localState`, but also subscribes to its change
- change will trigger component re-render

grabbou: Implemented as per above + Discord messages: https://github.com/final-ui/final/pull/76/commits/5e125232810763819d748ef88f8d43b2f6489ae4

packages/react/src/streams.ts
```diff
@@ -30,10 +24,19 @@ export function createWritableStream<S extends object>(
       }
     },
   }
-  return [writeState, stream]
+  return [
+    function writeState(updater) {
```
ericvicenti, line 28: lost the type signature here?

packages/react/src/state.tsx
```diff
@@ -115,4 +134,32 @@ export const increment = (value: number): IncrementStateModifier => ({
...
+      const subscription = stream.subscribe(() => {
+        forceUpdate({})
+      })
+      subscriptions.current.set(state.key, subscription)
+    }
+
+    return stream.get()
+  }
+
+  useEffect(() => {
```
ericvicenti, line 156: hmm, one code smell is that usually the subscription will happen in the useEffect that gets cleaned up in the return value of useEffect. like, does this even work with the double effect thing that React does in dev mode?

ericvicenti (commented):
this is complex but I don't see anything super hacky after a quick glance

will try to take a closer look later tonight

packages/react/src/state.tsx
grabbou: Interesting! I've been mostly testing this in unit tests where this does not apply of course. Good catch, we need to handle it!

grabbou: Fixed in https://github.com/final-ui/final/pull/76/commits/3d0f4f972478972ab697752c928006d1bd8bca8b

packages/react/src/state.tsx
```diff
@@ -115,4 +134,43 @@ export const increment = (value: number): IncrementStateModifier => ({
   value,
 })
 
-export const LocalState = createContext<LocalState>({})
+export const useLocalStateValues = () => {
+  const localState = useContext(LocalState)
+
+  const [, forceUpdate] = useState({})
+
+  const streams = useRef(new Map<string, Stream<JSONValue>>())
```
ericvicenti, line 142: if the state dependencies change, I guess this does not ever clean up the unused streams?

Because `get` for all states should happen during every render, maybe (maybe!) it makes sense to do this:
```
  const streams = useRef(new Map<string, Stream<JSONValue>>())
  streams.current = new Map()
```

ericvicenti (approved):
overall it looks good!

packages/react/src/state.tsx
grabbou: Going to turn this into an issue to address later. Now that I think about it, it's unlikely state dependencies change (at least within kit package), so while it's possible to support this, I can't think of a nice and easy unit test case to test it.

PR #78, opened 2024-06-04
fix: various typescript errors and better performance of recursive helpers

grabbou:
Fixing some annoying TypeScript errors in order to make it all work nicely locally and for users. The project builds fine at this stage.

Fixes explained line by line.

### DIFF
--- a/packages/playground/src/data-boundary.tsx
+++ b/packages/playground/src/data-boundary.tsx
@@ -1,4 +1,4 @@
-import { useStream } from '@final-ui/react'
+import { DataSource, useStream } from '@final-ui/react'
 import { WebSocketDataSource } from '@final-ui/ws-client'
 import { AlertCircle } from '@tamagui/lucide-icons'
 import React, { PropsWithChildren } from 'react'
@@ -9,8 +9,8 @@ export function DataBoundary({
   dataSource,
   path,
   children,
-}: PropsWithChildren<{ dataSource: WebSocketDataSource; path: string }>) {
-  const state = useStream(dataSource.state)
+}: PropsWithChildren<{ dataSource: WebSocketDataSource | DataSource; path: string }>) {
+  const state = useStream('state' in dataSource ? dataSource.state : undefined)
   const data = useStream(dataSource.get(path))
 
   if (data !== undefined) {
--- a/packages/react/src/jsx.tsx
+++ b/packages/react/src/jsx.tsx
@@ -5,7 +5,6 @@ import {
   ActionDataState,
   ComponentDataState,
   isComponentDataState,
-  JSONValue,
   ReferencedDataState,
   ServerDataState,
   ServerEventDataState,
--- a/packages/react/src/jsx.tsx
+++ b/packages/react/src/jsx.tsx
@@ -4,7 +4,6 @@ import { event } from './events'
 import {
   ActionDataState,
   ComponentDataState,
-  DataState,
   isComponentDataState,
   JSONValue,
   ReferencedDataState,
@@ -13,19 +12,21 @@ import {
   StateDataState,
 } from './template'
 
-// tbd: refactor and unify this
-type AllowedDataStates =
+type Props = Record<
+  string,
+  | JSONValue
   | ServerEventDataState
   | ActionDataState
   | ReferencedDataState
   | StateDataState
   | ((args: any) => any)
-
-type Props = Record<string, JSONValue | AllowedDataStates> & {
-  children: DataState
+> & {
+  children: ServerDataState
 }
 
-export type UI = ReactElement<Props> | ComponentDataState<ServerEventDataState>
+type ServerComponent = ComponentDataState<ServerEventDataState>
+
+export type UI = ReactElement<Props> | ServerComponent
 
 export const jsxs = jsx
 
@@ -33,7 +34,7 @@ export function jsx(
   componentFactory: (props: any) => UI,
   { children, ...passedProps }: Props,
   key?: string
-): ComponentDataState<ServerDataState> {
+): ServerComponent {
   const el = componentFactory(passedProps)
   if (isComponentDataState(el)) {
     return el
--- a/packages/react/src/jsx.tsx
+++ b/packages/react/src/jsx.tsx
@@ -12,21 +12,12 @@ import {
   StateDataState,
 } from './template'
 
-type Props = Record<
-  string,
-  | JSONValue
-  | ServerEventDataState
-  | ActionDataState
-  | ReferencedDataState
-  | StateDataState
-  | ((args: any) => any)
-> & {
-  children: ServerDataState
-}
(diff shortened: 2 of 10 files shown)


patches/@tamagui+progress+1.94.4.patch
```diff
@@ -0,0 +1,11 @@
+diff --git a/node_modules/@tamagui/progress/types/Progress.d.ts b/node_modules/@tamagui/progress/types/Progress.d.ts
+index 58f3db4..cbdfac4 100644
+--- a/node_modules/@tamagui/progress/types/Progress.d.ts
++++ b/node_modules/@tamagui/progress/types/Progress.d.ts
+@@ -254,5 +254,5 @@ declare const Progress: React.ForwardRefExoticComponent<Omit<import("@tamagui/co
+     }, import("@tamagui/core").StaticConfigPublic>;
+ };
+ export { createProgressScope, Progress, ProgressIndicator };
+-export type { ProgressProps, ProgressIndicatorProps };
++export type { ProgressProps, ProgressExtraProps, ProgressIndicatorProps };
```
grabbou, line 10: Temporarily using `patch-package` but hoping to send PR to Tamagui. W/o "ExtraProps" exported, it is not possible to re-export Tamagui-derived type with "declaration: true".

Issue that is encountered: https://github.com/microsoft/TypeScript/issues/5711

packages/playground/src/data-boundary.tsx
```diff
@@ -9,8 +9,8 @@ export function DataBoundary({
   dataSource,
   path,
   children,
-}: PropsWithChildren<{ dataSource: WebSocketDataSource; path: string }>) {
-  const state = useStream(dataSource.state)
+}: PropsWithChildren<{ dataSource: WebSocketDataSource | DataSource; path: string }>) {
+  const state = useStream('state' in dataSource ? dataSource.state : undefined)
```
grabbou, line 13: fix: typescript error when passing HTTPDataSource to Boundary.

The extra conditional is just to make TypeScript happy. `useStream` handles undefined already. That's why previous code was working just fine.

Unfortunately, `dataSource?.state` is going to fail too.

packages/react/src/jsx.tsx
```diff
@@ -13,27 +12,20 @@ import {
...
-  | ActionDataState
-  | ReferencedDataState
-  | StateDataState
-  | ((args: any) => any)
+export type UI = ReactElement<Props> | ServerComponent
 
-type Props = Record<string, JSONValue | AllowedDataStates> & {
-  children: DataState
+type ServerComponent = ComponentDataState<ServerEventDataState>
+type Props = ServerComponent['props'] & {
```
grabbou, line 18: refactor: do not duplicate "prop" types that can be defined for components on the server. Instead, reference props from the real type.

packages/react/src/template.tsx
```diff
@@ -13,28 +13,27 @@ export type ComponentDefinition<T extends Record<string, JSONValue>> = {
 }
 
 /** Data state */
-export type DataState<T = never> =
+export type DataState<T = EventDataState> =
```
grabbou, line 16: fix: ServerDataState (DataState on the server, before serialisation) now will not have EventDataState. This is correct. EventDataState without function on the server is going to result in an error, so should be properly typed. I will improve this type in a follow-up anyway.

packages/react/src/template.tsx
```diff
@@ -13,28 +13,27 @@ export type ComponentDefinition<T extends Record<string, JSONValue>> = {
...
-export type ComponentDataState<T = never> = {
+export type ComponentDataState<T = EventDataState> = {
   $: 'component'
   key?: string
   component: ComponentIdentifier
   children?: DataState<T>
   props?: Record<string, DataState<T>>
 }
-type ReferencedComponentDataState<T = never> = ComponentDataState<T> & {
+type ReferencedComponentDataState<T = undefined> = ComponentDataState<T> & {
```
grabbou, line 36: fix: by turning this into `undefined`, we can skip <T> when we want the default.

packages/react/src/utils.ts
```diff
@@ -1,4 +1,4 @@
-export function lookupValue(value: Record<string, any>, ref: (string | number)[]) {
+export function lookupValue(value: unknown, ref: (string | number)[]) {
```
grabbou, line 1: fix: we accept any value and we check for different types inside (we also have unit tests for it)

packages/ws-client/src/data-source.ts
```diff
@@ -34,7 +34,7 @@ export type UpdateWebsocketMessage = {
 export type EventResponseWebsocketMessage = {
   $: 'evt-res'
   key: string
-  res: ServerResponseDataState
+  res: ResponseDataState
```
grabbou, line 37: fix: the type on the client (after json.stringify) is obv. ResponseDataState (object w/o functions)

packages/react/src/utils.ts
ericvicenti: ok great, relying on inference here

packages/react/src/template.tsx
```diff
@@ -13,28 +13,27 @@ export type ComponentDefinition<T extends Record<string, JSONValue>> = {
...
-  | EventDataState
   | StateDataState
   | { [key: string]: DataState<T>; $?: never }
   | DataState[]
   | T
 /** Server data state */
 export type ServerDataState = DataState<ServerEventDataState>
 
-export type ComponentDataState<T = never> = {
+export type ComponentDataState<T = EventDataState> = {
```
ericvicenti, line 29: do we want to use `extends EventDataState` here as well?

packages/react/src/template.tsx
ericvicenti: hmm, when is the referenced value supposed to be undefined? Not sure I understand this part

packages/react/src/jsx.tsx
```diff
@@ -58,15 +58,24 @@ export function jsx(
   }
 }
 
-type Extend<P, K> = {
-  [Key in keyof P]: P[Key] extends infer T ? (T extends object ? K | Extend<T, K> : T | K) : never
-}
```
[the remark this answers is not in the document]
ericvicenti: Ah man I used to run into this all the time on zerve when I used super heavy TS inference

ericvicenti (approved):
Man, library TS is tricky, I barely understand 😂

Bummer we have to use patch-package especially for a peerDep which is scary, but hopefully it doesn't last long

packages/react/src/template.tsx
grabbou: LOL I thought this was working but it's obviously not. I wanted to make `ReferencedComponentDataState` completely optional but accidentally ended up passing `undefined` to ComponentDataState. Fixed

grabbou:
Yeah @ericvicenti I already spoke to Nate about this and hopefully he ships a fix soon!

PR #79, opened 2024-06-05
feat: rewrite kit to be (mostly) server-side, new state modifiers, few errors fixed

### LINKED ISSUE #31: example: Additional demo experiences for the Rise Mobile
Apart from a currently being developed "inventory" example, we also want to do the following:
- [ ] showcase form capabilities by doing an "inspection report" screen with various inputs as well as photo upload capabilities
- [ ] TBD

### LINKED ISSUE #44: refactor: RiseDropdownButton and RiseSelectField should be simplified/removed or moved to server-only utility function
To be honest, I don't think we need these components.

`DropdownButton` does not have any client-side state, so it makes no sense for us to maintain it. You can 100% construct it on the server without any limitations, but with greater flexibility. The way they're made right now makes it quite limiting, when it comes to deciding what do you want to render inside a Sheet component.

One could create a reusable component on the server (or a function) to reduce boilerplate in case they want to render in multiple places and want to avoid code repetition.

We could consider creating some server-side utilities as well. So, instead of *you* having to bundle rise-ui/kit in your mobile application, effectively increasing bundle size, you would import those components 100% on the server.

grabbou:
Fixes #44 
Fixes #31 

In this PR, we're finally rewriting `kit` to benefit from latest additions to this library.

Fixed few issues down the road that became evident once we started rewriting kit. 
Added relevant unit tests for each. 
Created new example.

There are still a few tweaks to be made and some features missing, but this PR is already big enough so we will do the remaining in follow-ups.

Note: I also removed: LongPressBottomSheet component, as it was a workaround on top of 3rd party Bottom Sheet library. We have fixed underlying issue in other commit and we can now further leverage Tamagui. We can consider adding this later, but users should be able to easily reference it themselves too.

Note: We should look into DropdownButton and SelectField and decide whether we need both, or just one.
We should decide if we need SelectField/SliderFields at all, vs using Tamagui directly. There's no added benefit within kit, and also, we're making it harder to style (with "Label" passed as string)


example/demo/src/inventory/ui.tsx
```diff
@@ -31,8 +32,12 @@ export function InventoryExample(ctx: UIContext) {
 function HomeScreen() {
   return (
     <YStack backgroundColor={'$background'}>
+      <RiseDropdownButton
+        button={<H2>Inventory</H2>}
+        options={inventory.map((item) => ({ key: item.key, label: item.title }))}
+      />
       <ScrollView contentContainerStyle={{ paddingBottom: 20 }}>
-        {inventory.map((item, idx) => (
+        {inventory.slice(0, 1).map((item, idx) => (
```
grabbou, line 40: We need to come up with better Example to show this button in action

grabbou:
Relevant idea #80 as I am going through this PR.

packages/kit/src/DropdownButton.tsx
```diff
@@ -1,70 +1,59 @@
...
+type Props = {
+  onSelect?: (item: string) => void
+  button?: JSX.Element
+  options?: {
+    key: Only<string>
+    label: string
+  }[]
+}
+
+export function DropdownButton(props: WithServerProps<Props>) {
```
grabbou, line 22: Previously, we would write:
```
export const RiseSwitchField = createComponentDefinition<typeof SwitchField>('RiseSwitchField')
```
to create a server component definition that can be used with JSX on the server.

Now, that we write the server component from scratch, we need a way to decorate its incoming props too. This is needed so that users can pass e.g. "StateDataState" where type is "string".

Example:

Consider the DropdownButton:
```
<DropdownButton onSelect={() => {}} />
```

in this case, `onSelect` is a function. However, you should be also able to write:
```
<DropdownButton onSelect={event(() => {}, { timeout: 1000 })} />
```

Same for e.g. `label`:
```
<DropdownButton options={[{ label: 'option-a' }]} />
```

can be `state`:
```
const label = state('my-label')
<DropdownButton options={[{ label }]} />
```

In practice, this just adds more type safety, but there's nothing required from user side. If you pass e.g. `label` prop straight to Tamagui (imported from `@final-ui/tamagui/server`), its type is already updated to accept `string | StateDataState` so there is no issue.

packages/kit/src/DropdownButton.tsx
```diff
@@ -1,70 +1,59 @@
...
               return (
                 <Button
                   key={item.key}
-                  onPress={() => {
-                    setOpen(false)
-                    props.onSelect?.(item.key)
-                  }}
+                  onPress={
+                    props.onSelect
+                      ? extend(props.onSelect, setStateAction(isOpen, false))
```
grabbou, line 46: I have created this little helper so that we can execute an additional action on top of already provided prop. Again, helper is needed, because `props.onSelect` can be either function, action, array of actions or an EventDataState.

Maybe we can have a better abstraction in the future?

packages/kit/src/SelectField.tsx
```diff
@@ -1,163 +1,113 @@
...
-            width="100%"
+            width="$3"
             height="$3"
           >
-            <YStack zIndex={10}>
-              <ChevronDown size={20} />
-            </YStack>
-            <LinearGradient
+            <YStack zIndex={10}>{/* <ChevronDown size={20} /> */}</YStack>
+            {/* <LinearGradient
```
grabbou, line 101: Need to bring back LinearGradient and ChevronDown (and other icons)

packages/kit/src/DropdownButton.tsx
```diff
@@ -1,70 +1,59 @@
...
+} from '@final-ui/tamagui/server'
 
-export function DropdownButton(props: z.infer<typeof DropdownButtonProps>) {
-  const { options } = props
-  const [open, setOpen] = React.useState(false)
+type Props = {
+  onSelect?: (item: string) => void
+  button?: JSX.Element
+  options?: {
+    key: Only<string>
```
grabbou, line 17: We automatically extend each `string` to also accept `StateDataState` since we can resolve state on the client down to primitive value.

However, `key` is special and there might be other special properties in the future that do not require/should not be anything but primitive value.

On one hand, we could consider adding support for `key` to be a `StateDataState` (in `template.tsx`, wrap `stateNode.key` with `renderProp` function call), but I find this out of scope.

This little helper is just to annotate type and is used within `WithServerProps` utility to detect which types should be extended.

packages/react/src/jsx.tsx
```diff
@@ -49,24 +48,31 @@ export function jsx(
...
+export type _Extend<T> =
+  T extends Only<infer U>
+    ? U
+    : T extends (...args: any) => any
+      ? T | ServerHandlerDataState
+      : T extends Array<infer U>
+        ? _DeepPartialArray<U>
+        : T extends object
+          ? WithServerProps<T>
+          : T extends null | undefined
```
grabbou, line 65: improvement: if prop is null or undefined, leave it as is. Previously, it would turn it into "StateDataState | null", so if there was the following type:
```
type Props = {
  onPress?: Function
}
```
it would turn it into
```
{
   onPress: Function | ServerHandlerDataState | ReferencedDataState | StateDataState | undefined
}
```
where in fact, we want
```
{
   onPress: Function | ServerHandlerDataState | undefined
}
```

packages/react/src/template.tsx
```diff
@@ -18,13 +18,14 @@ export type DataState<T = EventDataState> =
...
-  | ActionDataState
+  | HandlerDataState<T>
   | StateDataState
   | { [key: string]: DataState<T>; $?: never }
-  | DataState[]
+  | DataState<T>[]
   | T
 /** Server data state */
 export type ServerDataState = DataState<ServerEventDataState>
+export type ServerHandlerDataState = HandlerDataState<ServerEventDataState>
```
grabbou, line 28: This is just all types that are valid to be passed where component expects a function.

packages/tamagui/src/index.tsx
```diff
@@ -120,9 +129,18 @@ export const TamaguiComponents: ComponentRegistry = {
...
+    component: t.Select.ItemIndicator,
+  },
   SelectSheet: {
     component: t.Select.Sheet,
   },
+  SelectContent: {
+    component: t.Select.Content,
+  },
+  SelectValue: {
+    component: t.Select.Value,
```
grabbou, line 142: chore: Some components were missing. Might be more, as Tamagui is quite large. We will be monitoring.

packages/kit/src/DropdownButton.tsx
```diff
@@ -1,70 +1,56 @@
...
-            {options?.map((item) => {
+            {props.options?.map((item) => {
               return (
                 <Button
                   key={item.key}
-                  onPress={() => {
-                    setOpen(false)
-                    props.onSelect?.(item.key)
-                  }}
+                  onPress={[setStateAction(props.value, item.key), setStateAction(isOpen, false)]}
```
grabbou, line 45: This is not ideal (we should accept prop instead), but currently we do not have a way to change `event.payload`. `onPress` does not have any arguments, so `props.onSelect` will not work. It works for `SelectField`, because `onValueChange` already calls with the current value.

We may need to create at some point later a way to write the following:
```
onPress{bind(props.onSelect, item.key)}
```
or something.

packages/react/src/events.ts
```diff
@@ -3,21 +3,29 @@ import {
...
+  ServerHandlerDataState,
   StateDataState,
 } from './template'
 
 export function isServerEventDataState(obj: any): obj is ServerEventDataState {
   return isEventDataState(obj) && 'handler' in obj && typeof obj.handler === 'function'
 }
 
-export function event<T>(
+export function event<T extends any[]>(
```
grabbou, line 14: This makes sure that when you pass `event()` to a prop that has ServerEventDataState<[argA, argB]> signature, event(func) -> func will have correct type.

Screenshot_2024-06-05_at_20 08 22

packages/react/src/events.ts
```diff
@@ -3,21 +3,29 @@ import {
...
+): ServerEventDataState<T>
+export function event<T>(
+  func: HandlerFunction<[T]>,
   opts?: {
     actions?: ActionDataState[]
     timeout?: number
     args?: { [K in keyof T]: StateDataState<T[K]> }
   }
-): ServerEventDataState {
+): ServerEventDataState<[T]> {
```
grabbou, line 28: When you specify "args", payload is overwritten and you receive just one argument - your state values (as an object).

packages/react/src/template.tsx
```diff
@@ -58,15 +61,17 @@ export type ResponseDataState = {
   actions: ActionDataState[]
 }
 export type HandlerReturnType = ResponseDataState | JSONValue | void
-export type HandlerFunction<T = any> = (args: T) => Promise<HandlerReturnType> | HandlerReturnType
+export type HandlerFunction<T extends any[] = any[]> = (
+  ...args: T
```
grabbou, line 65: We're changing payload signature from single value to an array of values, so that if function takes multiple arguments, we can pass them all to the server.

packages/http-server/src/data-source.ts
```diff
@@ -62,7 +62,7 @@ export function createHTTPDataSource() {
               `Missing event handler on the server for event: ${JSON.stringify(message.event)}`
             )
           }
-          let res = await value.handler(payload)
+          let res = await value.handler(...payload)
```
grabbou, line 65: We now support an array of arguments to the handler, so components with multiple arguments (such as Tamagui ones) will work.

packages/kit/src/Form.tsx
```diff
@@ -1,91 +0,0 @@
-import { ComponentProps, createContext, ReactNode, useContext, useEffect, useState } from 'react'
```
grabbou, line 1: Form doesn't need any special components now. We support it out of the box and can consider adding it to the kit if manual impl. as a part of example turns out to be too complex.

packages/react/src/template.tsx
```diff
@@ -259,7 +263,13 @@ export function BaseTemplate({
       )
     }
     if (isStateDataState(propValue)) {
-      return render(getLocalStateValue(propValue), path)
```
grabbou, line 262: bug: this would previously throw an error if state was an object. we should call "renderProp" not "render" so that if it's an object or an array (e.g. nested with other state) this will work.

packages/kit/src/client/DraggableFlatList.tsx
```diff
@@ -0,0 +1,57 @@
+import DraggableFlatList, { ScaleDecorator } from 'react-native-draggable-flatlist'
```
grabbou, line 1: Everything that uses React and is client-side goes into this folder, otherwise into "server". Easier to differentiate that way.

packages/kit/src/index.tsx
```diff
@@ -1,61 +1,17 @@
...
-  },
-  RiseTextField: {
-    component: TextField,
-    validator: TextField.validate,
-  },
-  RiseSubmitButton: {
-    component: SubmitButton,
-    validator: SubmitButton.validate,
+  RNQRCode: {
+    component: QRCode,
```
ericvicenti, line 11: here's a funny thought, if we support SVGs we can probably render the QRCode on the server (although that might make the payload too large)

packages/kit/src/server/DropdownButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+  return (
+    <YStack>
+      <Button onPress={setStateAction(isOpen, true)}>{props.button}</Button>
+      <Sheet
+        forceRemoveScrollEnabled={isOpen}
+        modal={true}
+        open={isOpen}
+        onOpenChange={setStateAction(isOpen)}
+        dismissOnSnapToBottom
+        zIndex={100_000}
```
ericvicenti, line 41: 😅

packages/kit/src/server/SelectField.tsx
```diff
@@ -0,0 +1,122 @@
...
+              </SheetScrollView>
+            </SheetFrame>
+          </Sheet>
+        </Adapt>
+
+        <SelectTrigger>
+          <SelectValue placeholder={props.unselectedLabel} />
+        </SelectTrigger>
+
+        <SelectContent zIndex={200000}>
```
ericvicenti, line 62: these random large numbers are sending me 😂

packages/kit/src/server/SwitchField.tsx
```diff
@@ -0,0 +1,26 @@
...
+  label: string
+  onCheckedChange: (value: boolean) => void
+}
+
+export function SwitchField(props: WithServerProps<Props>) {
+  return (
+    <XStack alignItems="center" justifyContent="space-between">
+      <Label>{props.label}</Label>
+      <Switch
+        backgroundColor="#ddd"
```
ericvicenti, line 17: we should probably use a tamagui color token

packages/playground/src/provider/index.tsx
```diff
@@ -10,27 +9,21 @@ import { ToastViewport } from './ToastViewport'
 export function Provider({ children, ...rest }: TamaguiProviderProps) {
   const scheme = useColorScheme()
   return (
-    <BottomSheetModalProvider>
-      <TamaguiProvider
-        disableInjectCSS
-        defaultTheme={scheme === 'dark' ? 'dark' : 'light'}
-        {...rest}
+    <TamaguiProvider disableInjectCSS defaultTheme={scheme === 'dark' ? 'dark' : 'light'} {...rest}>
+      <ToastProvider
```
ericvicenti, line 13: I wonder if we should add a toast action

packages/react/src/__tests__/state.test.tsx
```diff
@@ -3,20 +3,32 @@ import React from 'react'
 
 import { DataSource, Template } from '../refs'
 import { response } from '../response'
-import { increment, setStateAction, state, toggle } from '../state'
+import { increment, lookup, setStateAction, state, toggle } from '../state'
```
ericvicenti, line 6: funny how we have `lookup` as well as `state` in the server side models. we should find an improved naming scheme

packages/react/src/jsx.tsx
```diff
@@ -49,24 +48,35 @@ export function jsx(
...
-  : T extends Array<infer U>
-    ? _DeepPartialArray<U>
-    : T extends object
-      ? Extend<T>
-      : T | ReferencedDataState | StateDataState
+export type WithServerProps<T> = { [P in keyof T]: _Extend<T[P]> }
+interface _ExtendArray<T> extends Array<_Extend<T>> {}
 
-export interface _DeepPartialArray<T> extends Array<_Extend<T>> {}
+export type _Extend<T> = T extends StateDataState
```
ericvicenti, line 61: thanks for explaining this to me earlier. we should probably write a big comment at some point that explains for newcomers to the code

packages/react/src/state.tsx
```diff
@@ -21,6 +26,11 @@ type IncrementStateModifier = {
   type: 'increment'
   value: number
 }
+type LookupStateModifier = {
```
ericvicenti, line 29: hmm, this one is interesting because it transforms the payload of the original event prop and passes it into state. different from the other state modifiers and maybe it should be `lookupPayload` or just `payload` with an optional lookup field on it

packages/tamagui/src/index.tsx
ericvicenti: yeah I wish we had an easy way to monitor Tamagui as more get added

packages/ws-server/src/data-source.ts
```diff
@@ -45,23 +46,27 @@ export function createWSServerDataSource() {
...
   async function get(key: string) {
-    let value = values.get(key)
-    if (typeof value === 'function') {
-      value = await value()
-    }
-    if (isReactElement(value)) {
-      throw new Error(
-        'Rise JSX not configured. You must set "jsx" to "react-jsx" and "jsxImportSource" to "@final-ui/react" in your tsconfig.json.'
-      )
+    if (!cache.has(key)) {
```
ericvicenti, line 57: hmm this change is what you mentioned in the voice message. we will have to work on this for the new model system, because its possible that it already caches

packages/ws-server/src/data-source.ts
```diff
@@ -45,23 +46,27 @@ export function createWSServerDataSource() {
...
+      }
+      if (isReactElement(value)) {
+        throw new Error(
+          'Rise JSX not configured. You must set "jsx" to "react-jsx" and "jsxImportSource" to "@final-ui/react" in your tsconfig.json.'
+        )
+      }
+      cache.set(key, value)
     }
-    return value
+    return cache.get(key)
```
ericvicenti, line 69: if a function with closure is defined I think its possible that data will get stale in the cache.. just something to keep an eye on

ericvicenti (approved):
🚀

packages/react/src/state.tsx
grabbou: I think `payload` with optional lookup field makes sense. I would name it `path`. And we can create `payload` a function!

packages/react/src/__tests__/state.test.tsx
grabbou: yeah!

packages/react/src/jsx.tsx
grabbou: we don't really have any documentation mechanism yet in place, ideally we can come up with something that allows us to write documentation inline alongside relevant code

packages/playground/src/provider/index.tsx
grabbou: definitely! and maybe it should take ref or component for custom toast. I will create an issue for this!

packages/kit/src/index.tsx
grabbou: Yeah, some random stats from the internet:
```
JPEG, 90%: 39.4 kB
JPEG, 50%: 25,7 kB
PNG: 870 Bytes
GIF: 6,8 kB
SVG: 29 kB
SVG, gzipped: 2,4 kB
```
PNG would be cool, actually!

packages/react/src/state.tsx
grabbou: renamed to `eventPayload` as we already had this name. Maybe it's a bit lengthy but we can always change.

grabbou:
Oops miss-click XD

PR #86, opened 2024-06-11
refactor: rename final to rise-tools

grabbou:
Updated Expo organisation to reflect new name as well. No other changes. We own npm organisation and packages will be pushed there at the next release.


ericvicenti (approved):
❤️

PR #87, opened 2024-06-11
feat: add support for toast action (initial)

### LINKED ISSUE #83: feat: create toast action
I wonder if we should add a toast action

_Originally posted by @ericvicenti in https://github.com/final-ui/final/pull/79#discussion_r1631438309_

grabbou:
Adds new action and lays some foundation work for future actions inside Rise. 

Fixes #83


apps/mobile/src/screens/connection.tsx
```diff
@@ -44,16 +52,21 @@ function ActiveConnectionScreen({ connection }: { connection: Connection }) {
...
         return true
       }
       if (name === 'navigate-back') {
         router.back()
         return true
       }
+      if (action.name[0] === 'toast') {
+        const { title, message } = (action as ShowToastAction).name[1]
+        toast.show(title, { message })
+        return true
```
grabbou, line 68: Given this code, I propose the following API change to action:

Variant 1:
```ts
type ActionDataState<T> = { $: 'action', name: string, payload: T }
```

Variant 2
```ts
type ActionDataState<T> = { $: 'action', payload: T }
```

In Variant 1, we assume all actions within the system will need to have `name` and `payload` (optional).

In Variant 2, we just rename `name` to `payload` as I find it strange to access values from `name`. The fact that first value of payload is string, and second value(s) are action-specific data is something Rise-specific. Users can adopt a different shape within their projects.

While Variant 2 gives more flexibility, I think Variant 1 might provide a bit more interoperability across different projects, in case users decide to create reusable components that provide support for some actions. 

For example, there could be a 3rd party "ToastAction" component that users can drop-in within their apps to provide toast capability. Right now, it's hardcoded and spread out in many places within Rise, but there's nothing stopping us from turning it into reusable component, just like "DataBoundary".

grabbou: After discussing this with @ericvicenti on Discord, we're going with:
```
type ActionDataState<N, T> = { $: 'action', name: string } & T
```
This type might be actually broken, but you get the idea. Action "extends" from base object and it can provide some additional information.
We're changing name to be string.

ericvicenti (approved):
Looks great. Can't wait for TS upgrade that infers automatically

grabbou:
Shame on me, I removed too much while doing refactor lately and broke navigation. Included hot-fix in this PR.

PR #88, opened 2024-06-12
WIP: finish removal of Solito and fix some bugs

grabbou:
Various fixes to the Rise UX:

- Fix form default values - by default, label is no longer `’’`, which means you must provide it (as per screenshot).
- After adding a new connection or editing existing one, we “goBack” in the hierarchy. Previously was `push`. Maybe it worked differently with `Solito`, but it does not make sense to push on new route
- Moved Expo Router specific hooks from “src” - they should reside in “app”
- Fixed transition - Expo Router will not push if you push same URL but with different path
- Better logic for NotFound error - use Redirect instead of for loop
- Support multiple nested paths with this new approach


grabbou:
Will re-open with smaller PRs.

PR #89, opened 2024-06-13
fix: empty screen on initial render

grabbou:


packages/react/src/refs.tsx
```diff
@@ -91,13 +91,10 @@ function findAllRefs(stateNode: DataState, dataValues: DataValues): Set<string>
...
 
 function createRefStateManager(
-  setDataValues: Dispatch<SetStateAction<DataValues>>,
+  [dataValues, setDataValues]: [DataValues, Dispatch<SetStateAction<DataValues>>],
   dataSource: DataSource,
   rootKey: string
 ) {
-  let dataValues: DataValues = {
-    [rootKey]: dataSource.get(rootKey).get(),
-  }
```
grabbou, line 100: Declaring `dataValues` inside `createRefStateManager` looks like a mistake. We already define `dataValues` as
```
const [dataValues, setDataValues] = useState<DataValues>({})
```
in the outer function.

Because the initial value is `{}` - empty object - it is expected that this value is populated on first render by resolving refs:
```
ensureSubscription(rootKey)
```

However, if `dataSource` already contains that value:
```
let dataValues: DataValues = {
    [rootKey]: dataSource.get(rootKey).get(),
  }
```

this will be e.g.
```
{ ui: { $: 'component' }}
```

As such, `setRefValue` will return `false` - did not set new value - because it's already there. 

However, since that `dataValues` is a different `dataValues` that we pass to `<Template />` component, its value will still be an empty object, resulting in lack of re-render. 

Solution: do not declare `dataValues` inside `createRefStateManager`, but read the value that is bound to `setDataValues` function.

grabbou: This sounds to me like a mistake, so just want to clarify there was no intended extra behavior that I missed here.

Original code: https://github.com/rise-tools/rise-tools/commit/34799708f7752a0458ebd944e0b9d7d4bfa34055#diff-61946fde60dcce67d9ba501090fc507ae95bc238c4efb58eab556807b4c4e196L82-L84

packages/react/src/refs.tsx
```diff
@@ -91,13 +91,11 @@ function findAllRefs(stateNode: DataState, dataValues: DataValues): Set<string>
 }
 
 function createRefStateManager(
+  dataValues: DataValues,
```
ericvicenti, line 94: this should work but I personally don't like reassigning arguments. I see them as a `const` so what I would do here is name this argument `initDataValues` then do `let dataValues = initDataValues` in the fn body

ericvicenti (approved):
Aside from my nitpick, this seems good!

packages/react/src/refs.tsx
grabbou: was going to ask you about this, but you beat me to it. that makes sense!

PR #90, opened 2024-06-13
refactor: remove remaining parts from solito

grabbou:
All works.

The only difference is that I removed `onLongPress` to edit connection. We have an icon that we render when `readonly` is `false`. Previously, it was possible to enter edit screen for a built-in connection which is not ideal.


grabbou:
We may need to add it later back, if we decide we're using it for web support after migrating to React Navigation. But for now, it's a bit confusing and potentially error prone to use this (that relies on React Navigation) with Expo Router, that says not to use React Navigation APIs.

ericvicenti (approved):
ok, I like the long press functionality but its not a big deal

but its nice to have a simpler setup without Solito

grabbou:
@ericvicenti would you like to have it back? No big deal - just need to add a check for long pressing on a read only connection.

ericvicenti:
Would be nice but it’s very low priority 

On June 14, 2024, GitHub ***@***.***> wrote:
> @ericvicenti <https://github.com/ericvicenti> would you like to have
> it back? No big deal - just need to add a check for long pressing on a
> read only connection.
>

—
Reply to this email directly, view it on GitHub
<https://github.com/rise-tools/rise-tools/pull/90#issuecomment-
2167538354>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AALKGTLSEON7RGQF6UOOY3DZHKTSTAVCNFSM6AAAAABJITGC2KVHI2DSMVQWIX3LMV43OSLTON2WKQ3PNVWWK3TUHMZDCNRXGUZTQMZVGQ>.
You are receiving this because you were mentioned.Message ID: <rise-
***@***.***>

PR #91, opened 2024-06-14
fix: "not found" route when fast refresh remounts the expo-router root, upgrade to Expo SDK 51

grabbou:
The bug was happening when Fast Refresh had to remount the Expo Router root and was resulting in "undefined" pathname / params, leading to "Not Found" screen.

I found the bug to be caused by our custom `<Stack />` definition:
```ts
<Stack>
    <Stack.Screen
      name="connection/[id]"
      // @ts-ignore
      getId={({ params }: { params: { id: string; path: string } }) => {
        return `${params.id}-${params.path}`
      }}
    />
</Stack>
```

Removing it fixed all errors. I bet it has something to do with the way `name="connection/[id]"` is declared. 

Anyways, **decided to rewrite from params to route params**, so that: a) we don't need custom getId() function anymore, as each screen is different, and b) we have nice interoperability with upcoming models PR and possibly, web routing.

State persistence now revives entire stack, not just last URL. This means, if you have screen hierarchy `Root, A, B, C`, it will reconstruct that exact hierarchy in development mode. previously, you would always end up with `Root, A`, because all subsequent screens only differentiated by `params` (`?path=xxx`) and we didn't save params in the storage. 

https://github.com/rise-tools/rise-tools/assets/2464966/35ce5433-835c-408c-aed3-932f1c18bfcc

**Other changes:**
- upgraded Expo SDK to 51 to ensure we are running latest Expo Router. Done with CLI, no changes required just deps suggested by Expo
- updated name of script "eject" to be "prebuild" and added `--clean` flag to clean-up native files when running. Note you must run this after merging this PR to get latest Rise client.
- before: `/connections/inventory?path=inventory:003:item`, now `/connections/inventory/inventory:003:item`
- it can also be nested, e.g. `/connections/inventory/items/003` in case we ever do it this way
- state persistence is now dev-only, meaning its behavior will have no impact on users in production


apps/mobile/app/_layout.tsx
```diff
@@ -15,22 +15,25 @@ export const unstable_settings = {
...
-  const router = useRouter()
+  const navigation = useNavigationContainerRef()
 
   useEffect(() => {
-    const route = storage.getString('navRoute')
-    if (route && route !== '/') {
-      setTimeout(() => {
-        router.push(route)
-      }, 500)
+    if (__DEV__) {
```
ericvicenti, line 21: why only dev mode? this is very useful in production as well so users can get back to the same location when they reopen the app

ericvicenti (commented):
Looks good. I thought there was a reason I didn't use path before but now I can't remember

apps/mobile/app/_layout.tsx
```diff
@@ -15,22 +15,25 @@ export const unstable_settings = {
...
+      const state = storage.getString('navRoute-state')
+      if (state) {
+        const [, ...routes] = JSON.parse(state)
+        setTimeout(() => {
+          for (const route of routes) {
+            navigation.dispatch(CommonActions.navigate(route))
+          }
+        }, 500)
```
ericvicenti, line 29: man this hack is so crazy. I wish we had real state persistence with Expo Router, and its so sad that we have an error when this value is much less than 500ms

apps/mobile/app/_layout.tsx
grabbou: good catch! others will be developing using Rise production build, so this is very helpful!

apps/mobile/app/_layout.tsx
grabbou: replaced with requestAnimationFrame tho we will have to revisit later with React Navigation.

PR #93, opened 2024-06-15
feat!: rename several APIs: Rise, ModelState

ericvicenti:


apps/mobile/src/screens/connection.tsx
```diff
@@ -24,25 +24,25 @@ const components = {
 }
 
 type RiseAction =
-  | ActionDataState<'navigate', { path: string }>
-  | ActionDataState<'navigate-back'>
-  | ActionDataState<'toast', { title: string; message?: string }>
+  | ActionModelState<'navigate', { path: string }>
+  | ActionModelState<'navigate-back'>
+  | ActionModelState<'toast', { title: string; message?: string }>
```
grabbou, line 29: As discussed, we really need to do a follow-up where we rename things that are not "ModelState". Probably only applies to: Event, Action, State (client-side) and Response.

grabbou (approved):
Looking good! Excited for the names unification. Not much to add here, just left some nits/comments to help you with follow-up documentation work!

PR #102, opened 2024-06-16
feat!: new API for defining actions

grabbou:
- Moves toast and router into separate packages
- Creates new API for definition actions


apps/mobile/src/screens/connection.tsx
```diff
@@ -22,45 +23,15 @@ const components = {
...
-  | ActionModelState<'toast', { title: string; message?: string }>
-
-function isRiseAction(action: ActionModelState): action is RiseAction {
-  return ['navigate', 'navigate-back', 'toast'].includes(action.name)
-}
-
 export function ConnectionScreen({ connection, path }: { connection: Connection; path?: string }) {
-  const toast = useToastController()
-  const router = useRouter()
+  const expoRouterActions = useExpoRouterActions({ prefix: `/connection/${connection.id}` })
```
grabbou, line 27: basePath might be a bit more idiomatic here

apps/mobile/package.json
```diff
@@ -14,6 +14,13 @@
     "@babel/runtime": "^7.18.9",
     "@expo/config-plugins": "~8.0.0",
     "@react-navigation/native": "^6.1.17",
+    "@rise-tools/react": "0.2.0",
```
grabbou, line 17: Some packages were missing here, so adding for completeness

packages/react/src/rise.tsx
```diff
@@ -56,6 +56,14 @@ export type ActionModelState<
   $: 'action'
   name: T
 } & K
+export type ActionModelStatePayload<T> = Omit<T, 'name' | '$'>
+export type ActionDefinition<
+  T extends ActionModelState<any, any>,
```
ericvicenti, line 61: why are there two generic arguments to ActionDefinition? seems like you need either T or K not both

also it is best practice to have a good name for the generic. the single letter thing is common but not very good for people trying to understand the code

ericvicenti (approved):
ok pretty good.

I also realized that users are required to set up toast manually. see `mobile/src/provider/index.tsx`. this stuff should be provided by the toast kit

grabbou:
> I also realized that users are required to set up toast manually. see mobile/src/provider/index.tsx. this stuff should be provided by the toast kit

In the morning, we discussed that we would rather leave setting up environment and required providers up to users, depending on their mobile application situation. For example, `kit-react-navigation` should expect React Navigation provider to be already in the tree. Expo Haptics should require being configured. 

We could make an exception for Toast, as it requires a bunch of "stuff" to be present, but as far as React Navigation / Expo Router / Expo Haptics are concerned, I would rather wait and see till we get more adoption.

packages/react/src/rise.tsx
grabbou: Second argument `K` is provided with default value so that I don't have to write:
```
export type ActionDefinition<
  T extends ActionModelState<any, any>
> = {
  action: (payload: ActionModelStatePayload<T>) => void
  validate?: (args: unknown) => ActionModelStatePayload<T>
}
```
but can move `ActionModelStatePayload<T>` into "variable".

ericvicenti:
yeah I think toast makes sense to include. of course navigation would require outside setup. not sure what configuration would be required for haptics

packages/react/src/rise.tsx
grabbou: > also it is best practice to have a good name for the generic. the single letter thing is common but not very good for people trying to understand the code

Yeah, I am in support for this change. I would say - maybe - makes sense to do a follow-up where we rename all Ts and Ks in the project to something more descriptive. Good opportunity to improve naming in general throughout the codebase (it's mostly just `rise.tsx` file where types live, so not that much of work)

ericvicenti: fine to do it in a followup!

and yes please lets just use T

grabbou: Fine! Applied - I think less letters, the better lol!

PR #104, opened 2024-06-17
feat: add kit-haptics, add validation to other kit action packages, better validation errors

### LINKED ISSUE #92: Rise/Kit: add actions for haptic feedback
Some buttons or server behavior should have haptic responses, should be pretty easy to support this

grabbou:
Fixes #92 

Validation:
- fix an error when validating action would always throw an error
- add better error message for invalid payloads
- add validators for all actions (in expo router, toast and haptics). In the future, this will be less boilerplate when we use `Typia` (can infer validator from TS type)
- Ensure validator (such as Zod) returns expected shape from `validate` function
- Slightly more convenient `ActionsDefinition` helper

### DIFF
--- a/docs/docs/kit/haptics.md
+++ b/docs/docs/kit/haptics.md
@@ -10,11 +10,11 @@ npm install @rise-tools/kit-haptics expo-haptics
 
 Shorthand for `haptics('impact')`
 
-- `haptics('impact', type?: 'heavy' | 'light' | 'medium' | 'rigid' | 'soft' = 'medium')`
+- `haptics('impact', type: 'heavy' | 'light' | 'medium' | 'rigid' | 'soft' = 'medium')`
 
 Use this action to provide haptic feedback for physical impact. Type is optional and defaults to `medium`.
 
-- `haptics('notification', type?: 'success' | 'failure' | 'error' = 'success')`
+- `haptics('notification', type: 'success' | 'failure' | 'error' = 'success')`
 
 Use this action to provide haptic feedback for success, failure, and warning. Type is optional and defaults to `success`.
 
--- a/docs/docs/kit/haptics.md
+++ b/docs/docs/kit/haptics.md
@@ -7,6 +7,21 @@ npm install @rise-tools/kit-haptics expo-haptics
 ## Actions
 
 - `haptics()`
-- `haptics('impact', type)`
-- `haptics('notification', type)`
+
+Shorthand for `haptics('impact')`
+
+- `haptics('impact', type?: 'heavy' | 'light' | 'medium' | 'rigid' | 'soft' = 'medium')`
+
+Use this action to provide haptic feedback for physical impact. Type is optional and defaults to `medium`.
+
+- `haptics('notification', type?: 'success' | 'failure' | 'error' = 'success')`
+
+Use this action to provide haptic feedback for success, failure, and warning. Type is optional and defaults to `success`.
+
 - `haptics('selection')`
+
+Use this action to let user know when a selection change has been registered.
+
+---
+
+Check [expo-haptics](Check https://docs.expo.dev/versions/latest/sdk/haptics/) documentation for more.
--- a/docs/docs/kit/haptics.md
+++ b/docs/docs/kit/haptics.md
@@ -3,3 +3,10 @@
 ```sh
 npm install @rise-tools/kit-haptics expo-haptics
 ```
+
+## Actions
+
+- `haptics`
+- `haptics('impact', type)`
+- `haptics('notification', type)`
+- `haptics('selection', type)`
--- a/docs/docs/kit/haptics.md
+++ b/docs/docs/kit/haptics.md
@@ -6,7 +6,7 @@ npm install @rise-tools/kit-haptics expo-haptics
 
 ## Actions
 
-- `haptics`
+- `haptics()`
 - `haptics('impact', type)`
 - `haptics('notification', type)`
-- `haptics('selection', type)`
+- `haptics('selection')`
--- a/example/demo/src/ui-controls/ui.tsx
+++ b/example/demo/src/ui-controls/ui.tsx
@@ -7,7 +7,8 @@ import {
 } from '@rise-tools/kit/server'
 import { navigate } from '@rise-tools/kit-expo-router/server'
 import { haptics } from '@rise-tools/kit-haptics/server'
-import { action, event, eventPayload, response, setStateAction, state } from '@rise-tools/react'
+import { toast } from '@rise-tools/kit-tamagui-toast/server'
+import { event, eventPayload, response, setStateAction, state } from '@rise-tools/react'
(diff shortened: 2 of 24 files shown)


packages/kit-tamagui-toast/src/index.ts
```diff
@@ -1,20 +1,24 @@
...
-export const useToastActions = (): ToastActions => {
+export const useToastActions = (): ActionsDefinition<ToastAction> => {
   const toast = useToastController()
 
   return {
     toast: {
       action: ({ title, ...options }) => {
         toast.show(title, options)
       },
+      validate: (payload) => ToastActionPayload.parse(payload),
```
grabbou, line 21: With `typia`, we will be able to write:
```
typia.assert<ActionPayload<ToastAction>>(input)
```
and call it a day!

grabbou:
Open questions:
- what do we want our example to be like?

TBD:
- I haven't tested it on actual device yet - should work just fine, but Simulator doesn't provide me feedback. No errors though.

packages/kit-haptics/src/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const NotificationActionPayload = z.object({
+  type: z.nativeEnum(Haptics.NotificationFeedbackType).optional(),
+})
+
+type HapticsActions = ActionsDefinition<ImpactAction | NotificationAction | SelectionAction>
+
+export const useHapticsActions = (): HapticsActions => {
+  return {
+    'haptics/impact': {
```
ericvicenti, line 19: I thought you wanted to use the package name as the prefix? So the action would be `@rise-tools/kit-haptics/impact`

packages/kit-tamagui-toast/src/index.ts
ericvicenti: cool, if the build step works! Also its a bit opaque to see the runtime code

packages/react/src/rise.tsx
```diff
@@ -57,10 +58,16 @@ export type ActionModelState<
   name: T
 } & K
 export type ActionModelStatePayload<T> = Omit<T, 'name' | '$'>
-export type ActionDefinition<T extends ActionModelState<any, any>> = {
-  action: (payload: ActionModelStatePayload<T>) => void
-  validate?: (args: unknown) => ActionModelStatePayload<T>
-}
+export type ActionsDefinition<ActionDefinition> = UnionToIntersection<
```
ericvicenti, line 61: before: I understood this type
after: I don't understand

not that it is wrong. just saying its hard to understand

example/demo/src/ui-controls/ui.tsx
```diff
@@ -185,3 +188,19 @@ function ShowToastExample() {
     </YStack>
   )
 }
+
+function HapticsExample() {
+  return (
+    <YStack>
+      <Button onPress={haptics()}>
```
ericvicenti, line 195: can we add buttons to test the rest of the haptic options?

docs/docs/kit/haptics.md
```diff
@@ -2,4 +2,11 @@
...
-```
\ No newline at end of file
+```
+
+## Actions
+
+- `haptics`
+- `haptics('impact', type)`
+- `haptics('notification', type)`
+- `haptics('selection', type)`
```
ericvicenti, line 12: I think selection doesn't have a type, does it?

Also we should specify the actual values. Feel free to just add some notes here for the valid values, and I will clean up the docs later

ericvicenti (approved):

packages/react/src/rise.tsx
grabbou: should we provide some documentation for it?

I agree it's complex. Not a huge fan of it. But it unlocks quite a bit of safety!

Let's suppose you have:
```
typeof action('navigate', { path }) // NavigateAction
typeof action('goBack') // GoBackAction
```
they are both ActionModelState. First one is ActionModelState<'navigate', { path: string }>, and second one is  ActionModelState<'goBack'>

When you create an object that is supposed to handle all different types of actions, you don't want to write:
```
const myActions = {
  'navigate': ActionPayload<NavigateAction>
}
```
because you're defining 'navigate' here and in case it changes in the other file where action helper is defined, you will have an unhandled action.

That's why I came up with this "helper" "ActionsDefinition" that takes union of all your actions.

You write ActionsDefinition<NavigateAction | GoBackAction> and it will do the following:

- When you pass union type to generic, it's basically "for each item in union type, apply this generic". So, ` ActionDefinition extends ActionModelState<infer Name, infer Action>` allows us to "unpack" Name and Action from each "ActionModelState" item in an union.
- From that point onwards, we just create a type that has "action" handler and "validate"

At this point `ActionsDefinition<NavigateAction | GoBackAction>` is converted into:

```
{ 'navigate': { action: (props) => void }} | {'goBack': { action: () => void }}
```

After wrapping everything with `UnionToIntersection`, we end up with a single object.

If there's
- argument mismatch
- wrong action name
- some actions are not handled
- extraneous actions are handled (no helper defined for such action or action was removed)
- payload type is different
- validator returns different type

You will receive type error.

grabbou: We can also return back to previous version, as it's more simple. Only challenge will be less type safety, but part of me feels like if we over engineer the types they will be holding us back really soon.

packages/kit-haptics/src/index.ts
grabbou: Yeah, I likely forgot about it, but it makes sense! 😆

packages/react/src/rise.tsx
grabbou: After reading my own explanation I realised that while added type safety is nice, I don't think that complexity is worth it.

packages/kit-haptics/src/index.ts
grabbou: Updated!

docs/docs/kit/haptics.md
grabbou: Yes, my bad. Updated `selection` type. I will cross-link to Expo for now and let you figure the best way to present it when you get to it!

example/demo/src/ui-controls/ui.tsx
grabbou: Yeah - so I will create button for each variant and group it in sections - one for impact, second one for notifications etc.

packages/kit-tamagui-toast/src/index.ts
grabbou: Yeah. For now, I am not continuing with it. I don't like patching TypeScript and it will not scale for users that ever run from source (and haven't patched their TypeScript). For things such as simple actions, we can probably write some basic if conditions ourselves. 

Anyways, it's a nice problem to have and I think we're aware of it, but it's definitely too early to look into it. Zod has nice DX and isn't going to be too hard to upgrade this way or another.

ericvicenti (approved):
😍

grabbou:
Was able to run on my iPhone and verified it works as expected.

PR #106, opened 2024-06-17
feat: example application for talk 

grabbou:
On top of example talk, this PR also:
- adds `<Theme />` tamagui component (I was planning to use it, but ended up not using it at this point)
- fixes an issue where rendering server defined function with JSX (e.g. `<Section />` component) would render without children

Here is what it looks like:

I know we talked a lot about keeping the implementation simple - that's why I tried to create helpers for some of the common layouts to keep it more semantic. Looking for feedback in this area!


example/demo/src/delivery/ui.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+function Survey() {
+  return (
+    <XStack backgroundColor="$green10" padding="$4" gap="$4" alignItems="center">
+      <LucideIcon icon="MessageCircleHeart" size="$6" color="white" />
+      <YStack justifyContent="flex-start" gap="$2">
+        <H3 color="white">Do you like the App?</H3>
+        <Button
+          backgroundColor="$green12"
+          pressStyle={{ backgroundColor: '$green11', borderColor: '$green11' }}
```
grabbou, line 75: Pretty sure there's better way with `<Theme />` to change color of Button w/o having to customise every single color. This will likely change. Or I will create ButtonPrimary and ButtonSecondary higher-order functions in this file.

packages/kit/src/index.tsx
```diff
@@ -12,6 +12,5 @@ export const RiseComponents = {
   },
   LucideIcon: {
     component: LucideIcon,
-    validator: LucideIcon.validate,
```
ericvicenti, line 15: 🥲

ericvicenti (approved):
Lets go!!

packages/kit/src/index.tsx
grabbou: Haha yeah, I had to let this one go for now. We need better validator for Icons for this and other Tamagui props (which is generally tricky, but we have it on the roadmap!)

grabbou:
Form is currently blocked by #113. So, I am going to merge it after adding some tiny little UI improvements, but revert last addition of a <Form /> component until its fixed.

ericvicenti:
Just some tiny nit-picky details for the visuals: Groceries,Restaurant,Taxi should be green to indicate they are server rendered, as we do in the slides

Also the padding above each header is too big

grabbou:
Gotcha! Let me tweak that! I definitely didn't check the keynote slides! Will do the tweaks!

PR #109, opened 2024-06-17
feat(jsx): fragments support

### LINKED ISSUE #65: feat: support JSX fragments in server code
Right now, they do not work. We need to map fragment on the client to React fragment.

grabbou:
Fixes #65 

- Automatically translates `<></>` to `<Fragment />`
- Exposes `<Fragment />` server-side component definition in case you want to use it explicitly via JSX


packages/react/src/jsx.tsx
```diff
@@ -86,3 +89,5 @@ export function createComponentDefinition<
 export function isReactElement(obj: any): obj is ReactElement {
   return obj !== null && typeof obj === 'object' && 'type' in obj && 'props' in obj && 'key' in obj
 }
+
+export const Fragment = createComponentDefinition<typeof React.Fragment>('Fragment')
```
grabbou, line 93: Shall we namespace this to `@rise-tools/react/Fragment`?

packages/react/src/jsx.tsx
```diff
@@ -10,20 +11,22 @@ import {
   StateModelState,
 } from './rise'
 
-export type UI = ReactElement<Props> | ServerComponent
```
grabbou, line 13: Clean up. This is no longer needed after `ws-server` and `http-server` were removed (via models PR). For some reason, two files were left hanging, so I deleted them too!

ericvicenti (approved):
looks good to me!

PR #110, opened 2024-06-18
fix: warnings about missing keys when children are static

grabbou:
So, now we know what "s" in "jsxs" stands for... 😆 [Static!](https://github.com/facebook/react/blob/main/packages/react/src/jsx/ReactJSX.js#L12-L13)

[The jsxs function indicates that the top array was created by React](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#detailed-design). 

In development, React uses "jsx" and "jsxs" factories to detect when array is created dynamically (e.g. via a loop) and keys are not provided. This is obviously better DX.

We are now introducing an advanced JSON field that will be set automatically when using JSX, based on the way you define arrays, to help you debug issues down the road and avoid render issues. If you're working with raw JSON, you should assume everything is dynamic and keys will be required on children.


packages/react/src/jsx.tsx
```diff
@@ -13,7 +13,14 @@ import {
 
 type ServerComponent = ComponentModelState<ServerEventModelState>
 
-export const jsxs = jsx
+// @ts-ignore
+export const jsxs = (...args) => {
+  // @ts-ignore
+  const d = jsx(...args)
+  // @ts-ignore
+  d.static = true
```
grabbou, line 21: We need to somehow annotate on the ComponentDataState that its children were created statically and as such, do not require keys to be present. 

Question: How should we transfer that information over the wire?

grabbou:
Note: React uses same factory in production for both static and dynamic. This is mostly DEV-only to improve debugging.

packages/react/src/rise.tsx
```diff
@@ -198,10 +199,17 @@ export function BaseRise({
           )
         }
       }
-      return (
-        <Component {...componentProps}>
-          {render(stateNode.children, [...path, 'children'])}
-        </Component>
+      // @ts-ignore
+      const jsxFactory = stateNode.static ? jsxs : jsx
```
grabbou, line 203: Note: [React uses same factory in production for both static and dynamic](https://github.com/facebook/react/blob/main/packages/react/src/jsx/ReactJSX.js#L20-L21). They only differentiate during development, likely to provide that key warning among other improvements.

It's probably better for us to still call static or dynamic selectively, in case that behavior changes in the future.

However, since same function is used in production right now, it leads to conclusion that differences if any are mostly dev specific.

grabbou: Alternatively, we could always call `jsxs`, no matter whether children are static or dynamic and that would trigger no warning on the client about missing keys. We would not add any additional properties to ComponentDataState.

This might be beneficial, because if you're using Rise Playground, you won't be able to see the logs and key warnings should appear on the server instead!

The only difference is during development when using this in-house as a part of your app. In that case you have access to logging and you should be able (or maybe not?) to see that!

grabbou:
Discussed we're going with first implementation and `$staticChildren` property. @ericvicenti

packages/react/src/jsx.tsx
```diff
@@ -12,14 +12,20 @@ import {
...
-
-export function jsx(
+type JSXFactory = (
   componentFactory: ((props: any) => ServerComponent | ReactElement) | undefined,
   { children, ...passedProps }: Record<string, any>,
   key?: string
-): ServerComponent {
+) => ServerComponent
+
+export const jsxs: JSXFactory = (...args) => {
```
grabbou, line 21: We generally prefer function declaration over arrow function as a top-level export, but this is meant to be used internally by React, so it's a fair tradeoff. The reason was to share `JSXFactory` type between these two w/o repeating.

ericvicenti (approved):

PR #114, opened 2024-06-18
refactor: rename `state` to `localStateExperimental` 

### LINKED ISSUE #82: refactor: move "state" to be experimental, rename to "localState" and remove random key to avoid conflicts on re-render
We generate random id for each state(). Http data source doesn't have cache - it will evaluate component each time (if it's a lazy function), which means the following won't work:
```
<RiseSlider
        value={[sliderValueMultiFirst, sliderValueMultiEnd]}
        onValueChange={[
          setStateAction(sliderValueMultiFirst, lookup([0, 0])),
          setStateAction(sliderValueMultiEnd, lookup([0, 1])),
        ]}
        onSlideEnd={async (_event, value) => {
          console.log('value', _event, value)
          // artificial stop
          await new Promise((resolve) => setTimeout(resolve, 2000))
          return response(null)
            .action(setStateAction(sliderValueMultiFirst, 25))
            .action(setStateAction(sliderValueMultiEnd, 60))
        }}
        min={0}
        max={100}
        step={1}
      />
```
Because the time `onSlideEnd` is run on the server in response to an action, will result in the outer function being run once again and hence, `state` having different IDs than the client.

grabbou:
Fixes #82 

As discussed during our f2f hackathon, local state is not our primary focus at this point, hence moving it to experimental. It generally works, but it has this limitation of requiring explicit key to be set. The API may change in the future as well as implementation.

However, we're keeping it there as its quite useful (in experimenting)

- [x] Rename state as experimental
- [x] Explicitly require unique key to be set (will be fixed at some point, but we have other priorities for now)

@ericvicenti, question: shall we create a simple "experimental" docs page for state? I know API may change, but we have quite a few nice abstractions already. Maybe it's worth having a skeleton just for our future selves.


ericvicenti (approved):

ericvicenti:
Lets not write the docs for state yet, maybe once we have finished all the other docs 😂

PR #115, opened 2024-06-19
feat!: simplify response() API

### LINKED ISSUE #63: chore: refactor response
Each server-side function should be able to:
1. Return JSONValue to automatically resolve with status 200 and that value
2. Return Action(s) to automatically resolve with status 200, empty response and actions array
3. Return `response(opts)` object that contains different options, including payload, actions and statusCode

grabbou:
Before this PR, your function handler could only return JSON Value. If you wanted to return an action, you would have to do this:
```
return response(null).action()
```

Fixes #63 

After this PR, you can do the following:
```
onPress={() => return { userName: "Mike" }} // return payload
onPress={() => return response(actions)} // return both payload and actions
onPress={() => return response(payload, { actions })} // return both payload and actions
```

This PR brings response API similar to `event()`


grabbou:
Also added some very basic documentation which needs to be rewritten 100%, but hopefully it helps!

docs/docs/guides/actions-events.md
```diff
@@ -1 +1,223 @@
...
+
+```tsx
+import { goBack } from "@rise-tools/kit-expo-router"
+import { haptics } from "@rise-tools/kit-haptics"
+
+function UI() {
+  return (
+    <Form
+      onSubmit={() => {
+        return [
```
ericvicenti, line 192: this behavior might be confusing. it looks like this would go back to the client with these events as the response.

I think we may require you to use `return response(something, { actions: [] })` here

ericvicenti (commented):
thanks for writing the doc here!

lots to improve on the docs still but its a good start

I'm not super confident that we want this behavior of returning actions from your event and magically detecting that, rather than having people use `response()` explicitly

docs/docs/guides/actions-events.md
grabbou: Yeah, I definitely agree, it might be confusing when both appear at the same time especially:
```ts
onSubmit={() => {
   if (success) return [action, action]
   return [{ error: 'Wrong name', field: 'name' }]
}
```

The challenge is that many times, you'll be just returning an action or multiple actions (e.g. in response to button press) and not returning any response:
```
return response(null, { actions: [] })
```

Feels like we could do better than that.

Do you have any ideas?

ericvicenti: idk, `responseActions([...])`

grabbou: not sure either. Let's keep this PR open for some time until we figure out a better API later today

grabbou:
- `{}`
- `response(action(s))`
- `response(payload, { action: action(s) })`

grabbou:
Done, also provided some additional type safety!

ericvicenti (approved):
this improved type safety is 🔥

packages/react/src/response.ts
```diff
@@ -1,26 +1,48 @@
...
-      return this
-    },
+    actions: opts?.actions,
   }
 }
 
-export type ServerResponseModelState = ResponseModelState & {
-  action(action: ActionModelState): ServerResponseModelState
-  status(code: number): ServerResponseModelState
+export function errorResponse<T>(...args: Parameters<typeof response<T>>): ResponseModelState<T> {
```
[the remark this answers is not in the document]
grabbou: We could (potentially), but we don't really use it in anyway on the client yet. Will turn this into issue, because honestly I don't have a ready strategy at this point!

PR #117, opened 2024-06-19
feat: add @rise-tools/kit-svg

### LINKED ISSUE #97: Rise/Kit: Support SVGs
The kit should have a component for rendering SVGs from the server

grabbou:
Fixes #97 

Adds `react-native-svg` supported by latest Expo SDK.

The example is quite ugly, but I just copied some examples from React Native SVG to demonstrate various functionality

- static SVG
- SVG from URI
- SVG as string

This is great opportunity to get creative and include some fun stuff, so let's get creative!

### DIFF
--- a/apps/mobile/src/screens/connection.tsx
+++ b/apps/mobile/src/screens/connection.tsx
@@ -1,6 +1,7 @@
 import { RiseComponents } from '@rise-tools/kit'
 import { ExpoRouterComponents, useExpoRouterActions } from '@rise-tools/kit-expo-router'
 import { useHapticsActions } from '@rise-tools/kit-haptics'
+import { SVGComponents } from '@rise-tools/kit-svg'
 import { useToastActions } from '@rise-tools/kit-tamagui-toast'
 import { Rise } from '@rise-tools/react'
 import { TamaguiComponents } from '@rise-tools/tamagui'
@@ -14,6 +15,7 @@ const components = {
   ...TamaguiComponents,
   ...RiseComponents,
   ...ExpoRouterComponents,
+  ...SVGComponents,
 }
 
 export function ConnectionScreen({ connection, path }: { connection: Connection; path?: string }) {
--- a/example/demo/src/ui-controls/ui.tsx
+++ b/example/demo/src/ui-controls/ui.tsx
@@ -7,7 +7,7 @@ import {
 } from '@rise-tools/kit/server'
 import { navigate, StackScreen } from '@rise-tools/kit-expo-router/server'
 import { haptics } from '@rise-tools/kit-haptics/server'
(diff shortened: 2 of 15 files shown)


example/demo/src/ui-controls/ui.tsx
```diff
@@ -50,11 +52,51 @@ function UI() {
...
+        <SvgCssUri
+          width="100"
+          height="100"
+          uri="https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/ruby.svg"
+          onError={toast('Failed to load SVG from URI')}
+        />
+      </YStack>
+      <YStack>
+        <H4>SVG string</H4>
+        <SvgCss
```
ericvicenti, line 81: calling this `SvgCss` feels off to me. the SVG here is XML so maybe it should be called SvgXml

Same with `SvgCssUri`. maybe should just be `SvgUri` (because people using a svg from a URL won't actually care that it is a xml)

apps/mobile/src/screens/connection.tsx
```diff
@@ -1,6 +1,7 @@
 import { RiseComponents } from '@rise-tools/kit'
```
ericvicenti, line 1: I guess we want to eventually remove all of these `kit` components because this represents the ones that are not organized in other packages. One day this would become the umbrella package

ericvicenti (approved):

apps/mobile/src/screens/connection.tsx
grabbou: Yup! Once we refactor dropdown (discussed in Discord) to a separate package with some better customization, this can be removed and replaced with umbrella!

example/demo/src/ui-controls/ui.tsx
grabbou: Yeah I haven't looked into it too much, just re-exported everything from SVG without doing anything extra. I find some of these makes quite verbose too, but that's the API. I would be skeptical about changing it too much to avoid confusion. But maybe we can build a higher order SVG component that does that?

grabbou: I guess here `SvgCss` is b/c there's `CSS` in that SVG string and they use different components for it, likely to optimise.

grabbou: It's `SvgXml` when there's no `CSS` inside. Should've been `SvgCssXml` to keep it consistent I guess but

grabbou: > Same with SvgCssUri. maybe should just be SvgUri (because people using a svg from a URL won't actually care that it is a xml)

All of these are available, depending on whether XML contains <style />

From uri:
- SvgUri  <- no CSS
- SvgCssUri  <- with CSS

From string:
- SvgXml <- no CSS
- SvgCss <- with CSS

PR #120, opened 2024-06-20
feat: add @rise-tools/kit-forms 

grabbou:
Adds new Form utilities on top of Tamagui, suffix: "Field". 
Also, moves SliderField, SwitchField and SelectField from "kit" to "kit-forms" (will be re-exported later anyways for clarity)
Groups all examples under single "Form"
Removes use of experimental local state as its no longer required to support forms in the app

Validation will be provided in a follow-up.


example/demo/src/delivery/ui.tsx
```diff
@@ -34,39 +18,11 @@ function UI() {
...
-    setName('')
-    setFeedback('')
-
-    // return response
-    return response(null).action(toast('Thank you for submitting your feedback')).action(goBack())
-  }
-  
   return (
-    <>
+    <YStack>
```
grabbou, line 23: I temporarily remove form from example until we figure it out!

example/demo/src/ui-controls/ui.tsx
```diff
@@ -56,94 +41,50 @@ function UI() {
...
+            .action(toast('Thank you for submitting your feedback'))
+            .action(goBack())
+        }}
+      >
+        <InputField id="name" label="Name" />
+        <TextField id="feedback" label="Feedback" />
+        <CheckboxField id="subscribe" label="Subscribe to newsletter" defaultChecked={true} />
+        <SliderField id="rating" label="Rating" defaultValue={[20, 60]} />
+        <SwitchField id="anonymous" label="I want to be anonymous" />
+        <SelectField
```
grabbou, line 59: Shall we keep the form example like this? @ericvicenti 

Or... 
- Input
- Input with custom label (you can render string or pass custom element)
- TextField
- Slider
- Range Slider
- Switch
- Toggle Group Horizontal
- Toggle Group Vertical
- Toggle Multiple
- Checkbox
- Radio Group

packages/kit-forms/src/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  SelectField,
+  SliderField,
+  SubmitButton,
+  SwitchField,
+  TextField,
+  ToggleGroupField,
+} from './form'
+
+export const FormComponents = {
+  '@rise-tools/kit-forms/RiseForm': {
```
ericvicenti, line 15: if we use this deep namespace we can just call this Form, no need for RiseForm

packages/kit-forms/src/server.ts
```diff
@@ -0,0 +1,34 @@
+import { createComponentDefinition } from '@rise-tools/react'
+
+import type * as Kit from './form'
+
+export const RiseForm = createComponentDefinition<typeof Kit.RiseForm>(
+  '@rise-tools/kit-forms/RiseForm'
```
ericvicenti, line 6: I have mixed feelings about the long names caused by this namespace pattern. good for safety and bad for anybody who has to manually write component definitions

maybe we should get rid of the `@` because it doesn't really provide any value

packages/kit-forms/src/index.tsx
grabbou: True! For the exported constant, I wanted to avoid collision with Tamagui form tho!

packages/kit-forms/src/server.ts
grabbou: BTW in the future we can probably auto generate this file as it's quite repetitive.

grabbou:
I have patched tamagui with `patch-package` and ignored one warning. Issues are either reported or pending PR merge/release in Tamagui. Created issue to track updates in the future and don't forget to get back to it!

example/demo/src/ui-controls/ui.tsx
grabbou: I am going to rewrite it to showcase all different types instead. Improvements are welcome!

Issue #72, opened 2024-06-01
dx: provide error message when passing mixed type as an array to function handler


grabbou:
You can only do `event()`, `function` or `action` or `action[]`. If you mix multiple types in an array, this will not work and should be signalled to the user.

TheKohan:
Hi, would happily start contributing here. Is this issue still up ? I've ran through the repo couple of times and I'm starting to get a hang of it (though everything is still foggy). Let me know if this is not to complex for a introductory PR.

grabbou:
Hey, yes, I think this is good one. We already had something similar in the past, so searching the repository might help! Feel free to join our Discord server - https://discord.gg/khS5RYhS5K!

grabbou:
Fixed in #124

PR #124, opened 2024-06-21
fix: throw error on mixed values array passed to `handlerFunction`

TheKohan:
Hi, 

Added detection of mixed type arrays as specified in #72. Created one test suite to cover this case.


grabbou:
Done! Apologies for the delay, I owe you coffee! Meanwhile, let's get moving 🔥

PR #126, opened 2024-06-21
fix: set static children with jsx at the right level when using server-side higher order functions

grabbou:
> [!WARNING]
> JSX support is still experimental and warnings, bugs are expected. This is just one of them and I guess not the last one!

Changelog:

- Rename `jsx` to `jsx-runtime` (default file name expected by Babel / TypeScript for custom JSX factories)
- Use JSX in unit test, instead of constructing it manually (this is crucial, b/c we now have jsx and jsxs)
- Provide a new unit test that ensures the bug is fixed

What was the root problem?

When we have the following:
```tsx
const Section = ({ children }) => {
  return (
    <View>
      <View>{children}</View>
    </View>
  )
}
const Home = () => (
  <Section>
    <View>Foo</View>
    <View>Bar</View>
  </Section>
)
```

This will be converted into:
```tsx
import { jsx, jsxs } from '@rise-tools/react/jsx-runtime'
const Section = ({ children }) => {
  return jsx(View, {
    children: jsx(View, {
      children,
    }),
  })
}
const Home = () =>
  jsxs(Section, {
    children: [
      jsx(View, {
        children: 'Foo',
      }),
      jsx(View, {
        children: 'Bar',
      }),
    ],
  })
```

Our current implementation (as below)
```tsx
export const jsxs: JSXFactory = (...args) => {
  return {
    ...jsx(...args),
    $staticChildren: true,
  }
}
```
will always add `$staticChildren` to the result of `jsx` call.

However, this does not work when intermediate component changes the structure. 

Let's look at the snippet shared earlier to break it down:
```tsx
const Home = () =>
  jsxs(Section, {
    children: [
      jsx(View, {
        children: 'Foo',
      }),
      jsx(View, {
        children: 'Bar',
      }),
    ],
  })
```

The return type of `jsxs` function call is not going to be:
```ts
{
  $: 'component',
  type: 'Section',
  $staticChildren: true,
  children: [...] // <-- this is not true
}
```

but:
```ts
{ 
   $: 'component',
   type: 'View',
   $staticChildren: true, // <-- $staticChildren should not be set here, but inside children
   children: { $: 'component', type: 'View', children: [...] } // <-- it's an object, result of Section() on the server
}
```

Because "Section" is server-side helper, it actually modifies the structure. As a result, children is no longer an array at the top-level, but single object, because "Section" returns single View, and then, renders passed children inside.

This will trigger warning that we tried to call `jsxs` on a single child (object). It will also trigger warning about missing keys, because static children were rendered few levels deeper but without `$staticChildren` set side by side.

Solution is to propagate a special property alongside children object and check for its presence right where we render it.


packages/react/src/jsx-runtime.tsx
```diff
@@ -18,11 +18,14 @@ type JSXFactory = (
   key?: string
 ) => ServerComponent
 
-export const jsxs: JSXFactory = (...args) => {
-  return {
-    ...jsx(...args),
-    $staticChildren: true,
-  }
+class StaticChildrenArray extends Array {}
```
grabbou, line 21: Is there better way to associate that information with an array? "children.$static = true"? "Object.defineProperty" & "non-enumerable"? 

Users can still write `children.map()` on the server, so our information should be opaque and should not affect results of `map`.

BTW, if you write `children.map()` on our `StaticChildrenArray`, it will covert to `Array` anyway which is good - children are no longer static.

packages/react/src/__tests__/jsx.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+    Object {
+      "$": "component",
+      "children": Object {
+        "$": "component",
+        "children": Array [
+          Object {
+            "$": "component",
+            "children": 0,
+            "component": "View",
+            "key": undefined,
```
grabbou, line 126: tip for the future, we could probably optimise `jsx` to strip undefined values to reduce payload.

packages/react/src/__tests__/jsx.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+  }
+  expect(
+    <Section>
+      <View>Foo</View>
+      <View>Bar</View>
+    </Section>
+  ).toMatchInlineSnapshot(`
+    Object {
+      "$": "component",
+      "children": Object {
```
grabbou, line 71: I have verified this test was failing before my change! `$staticChildren: true` was on this line, not on line 73

ericvicenti (approved):
great catch!

PR #127, opened 2024-06-21
feat: add "@rise-tools/kit-lucide-icons"

grabbou:
Moving components from "kit" to prepare it for being umbrella.


ericvicenti (approved):
awesome. maybe the example can showcase more of the icons, but unless we implement some sort of paginated request then we will run into performance issues if we try to render all of them at once

PR #129, opened 2024-06-22
feat: replace dropdownbutton with more universal bottomsheet

grabbou:
Adds missing trigger button (to open the sheet) and close button (to close it), so users can compose bottom sheet in similar fashion to other Tamagui components (inspired by Dialog.Trigger and Dialog.Close).

Sounds like good feature request to Tamagui to be honest!


packages/kit/src/client/BottomSheet.tsx
```diff
@@ -0,0 +1,73 @@
...
+          <Sheet.Frame padding="$4" justifyContent="center">
+            {children}
+          </Sheet.Frame>
+        </Sheet>
+      </YStack>
+    </BottomSheetContext.Provider>
+  )
+}
+
+export function BottomSheetTrigger({ children, ...props }: ButtonProps) {
```
ericvicenti, line 45: should this be named BottomSheetTriggerButton? or is the button obvious?

ericvicenti (approved):

PR #131, opened 2024-06-22
feat: rewrite DraggableFlatList + add RN FlatList

grabbou:
As per title.


packages/kit/src/client/DraggableFlatList.tsx
```diff
@@ -1,53 +1,43 @@
...
+      onDragEnd={(e) => {
+        setData(e.data)
+        props.onReorder?.(e.data.map((item) => item.key))
       }}
-      ListHeaderComponent={() => props.header}
-      ListFooterComponent={() => props.footer}
-      renderItem={(row) => {
-        const { item, drag, isActive } = row
+      renderItem={({ item, drag, isActive }) => {
+        const Item = () => item.label
```
grabbou, line 36: DraggableFlatList specific tweak -> https://github.com/computerjazz/react-native-draggable-flatlist/issues/496

packages/react/src/state.tsx
```diff
@@ -33,10 +33,7 @@ export function isStateUpdateAction<T extends JSONValue>(
   return action.name === 'state-update'
 }
 
-export function localStateExperimental<T extends JSONValue>(
-  initialValue: T,
-  key: string
-): StateModelState<T> {
+export function localStateExperimental<T>(initialValue: T, key: string): StateModelState<T> {
```
grabbou, line 36: I previously did this example using local state and these fixes were required. We're not using local state now, but will leave it there for the future! Basically, value can be anything, not just JSONValue. Otherwise, it was complaining when we included React Elements in the state.

docs/docs/kit/index.md
```diff
@@ -1,11 +1,11 @@
 # Rise Kit
 
-## @rise-tools/kit
+## @rise-tools/kitchen-sink
```
grabbou, line 3: In preparation for umbrella package! Also, created doc page for Kit itself

ericvicenti (approved):
nice!

PR #132, opened 2024-06-23
feat: create new umbrella package and use it instead

grabbou:
Two questions left


packages/kitchen-sink/src/index.ts
```diff
@@ -0,0 +1,10 @@
+export * from '@rise-tools/kit'
+export * from '@rise-tools/kit-expo-router'
+export * from '@rise-tools/kit-forms'
+export * from '@rise-tools/kit-haptics'
+export * from '@rise-tools/kit-linking'
+export * from '@rise-tools/kit-lucide-icons'
+export * from '@rise-tools/kit-qrcode'
+export * from '@rise-tools/kit-svg'
+export * from '@rise-tools/kit-tamagui-toast'
+export * from '@rise-tools/tamagui'
```
grabbou, line 10: I think we may want to rename `tamagui` to `kit-tamagui`?

packages/kitchen-sink/src/server.ts
```diff
@@ -0,0 +1,12 @@
...
+export * from '@rise-tools/kit-forms/server'
+export * from '@rise-tools/kit-haptics/server'
+export * from '@rise-tools/kit-linking/server'
+export * from '@rise-tools/kit-lucide-icons/server'
+export * from '@rise-tools/kit-qrcode/server'
+export * from '@rise-tools/kit-tamagui-toast/server'
+export * from '@rise-tools/tamagui/server'
+
+// tbd: "Circle" collides with Tamagui.Circle. Rename one of them or namespace?
+export * as SVG from '@rise-tools/kit-svg/server'
```
grabbou, line 12: Should we rename "Circle" inside SVG, "Circle" inside Tamagui, or leave it exported like this? (Kinda sucks to destruct on the client, b/c you have to write const { Circle } = SVG.

packages/kitchen-sink/src/index.ts
ericvicenti: yes I agree, lets do that

packages/kitchen-sink/src/server.ts
ericvicenti: ooh damn, maybe they should both be accessed through the namespace, otherwise people may accidentally import the wrong thing.

apps/mobile/src/screens/connection.tsx
```diff
@@ -1,14 +1,17 @@
...
+  FormComponents,
+  LucideIconsComponents,
+  QRCodeComponents,
+  SVGComponents,
+  TamaguiComponents,
+  useExpoRouterActions,
+  useHapticsActions,
+  useLinkingActions,
+  useToastActions,
+} from '@rise-tools/kitchen-sink'
```
ericvicenti, line 13: hmm at first I assumed it would be `import { Components, useActions } from kitchen sink` but now that I think about it, this is nicer because it is more clear

ericvicenti (approved):

apps/mobile/src/screens/connection.tsx
grabbou: yeah you can still add kitchen-sink as a dependency but because it's all peer dependencies, selectively install dependencies as you expand your functionality, which is nice!

PR #133, opened 2024-06-23
feat: drop "@" from namespace names, add namespace to tamagui package

### LINKED ISSUE #122: refactor: remove unnecessary "@" from the namespace
I have mixed feelings about the long names caused by this namespace pattern. good for safety and bad for anybody who has to manually write component definitions

maybe we should get rid of the `@` because it doesn't really provide any value

_Originally posted by @ericvicenti in https://github.com/rise-tools/rise-tools/pull/120#discussion_r1648516486_

grabbou:
Fixes #122 

I hope that in the future, we can auto-generate some of this (e.g. server component definitions, which is pretty repetitive). 

Meanwhile, making it SLIGHTLY shorter as per our previous discussion. Did namespace entire Tamagui package at that occasion, as it wasn't touched in a while.


ericvicenti (approved):
great I'm guessing everything is consistent now!

PR #135, opened 2024-06-23
fix: view model update when subscribed

ericvicenti:
View model was not firing subscription updates when the dependency changes. This was never implemented.

Using a timeout(1) to handle cases when two view dependencies change, we only want to trigger one update

### DIFF
--- a/packages/server/src/model-view.ts
+++ b/packages/server/src/model-view.ts
@@ -1,4 +1,3 @@
-
 import { getModelState } from './model-utils'
 import { ValueModel, ViewModel } from './types'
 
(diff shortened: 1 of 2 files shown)


ericvicenti:
fixes https://github.com/rise-tools/rise-tools/issues/134

packages/server/src/model-view.ts
```diff
@@ -22,6 +22,8 @@ export function view<T>(
           model,
           model.subscribe(() => {
             isInvalid = true
+            clearTimeout(updateSchedule)
+            updateSchedule = setTimeout(get, 1)
```
grabbou, line 26: why `1` tho? curious if you tried just `setTimeout(get)` and didn't work?

ericvicenti: it is the same behavior

ericvicenti: according to TS "When delay is larger than 2147483647 or less than 1, the delay will be set to 1. Non-integer delays are truncated to an integer."

grabbou: gotcha, definitely wasn't aware of that behavior!

PR #137, opened 2024-06-24
feat: better Rise UX

### LINKED ISSUE #107: feat(rise): better UX for connections in the app
Rather than push new screen, different transition and instead of back button, show something like "X" or connection switch.

grabbou:
Fixes #107
Relevant discussion #136 

Changelog:
- Each connection is now inside a nested navigator. That navigator is very portable and in the future, could be made available to the users as drop-in screen in their React Navigation routing
- Root connection screen is now presented without a header with a slide from bottom animation. To dismiss, there's new Rise icon in the top left corner that will take you back to the Home Screen
- Each connection has its own header from the nested navigator
- Fixed cut off Rise logo on the home screen due to incorrect aspect ratio
- Fixed `addConnection` not returning `id` as expected by the caller
- Created new React Navigation package with actions and exported it from Kitchen Sink instead of Expo Router. I did this because: a) we don't use Expo Router anywhere in Rise and b) goBack() and navigate() had a name conflict with Expo Router package. 
- State persistence is now dead simple and works with our nested structure!

Here's video demonstrating latest behavior:
https://github.com/rise-tools/rise-tools/assets/2464966/10f05099-4a71-421d-960b-6cd55eb02fba

**To do:**
- [x] Tamagui Toast appears under the modal
- [x] Verify everything works while refreshing the app
- [x] Test
- [x] Add deep linking for each screen (and also nested screens in connections)


apps/mobile/src/screens/connection.tsx
```diff
@@ -1,54 +1,96 @@
...
+import { useConnection } from '../connection'
 import { DataBoundary } from '../data-boundary'
 import { useModelSource } from '../model-sources'
+import { RootStackParamList } from '.'
+import { NotFoundScreen } from './not-found'
 
 const components = {
   ...TamaguiComponents,
   ...RiseComponents,
-  ...ExpoRouterComponents,
```
ericvicenti, line 24: looks like we need to replace this at some point

packages/kitchen-sink/package.json
```diff
@@ -37,12 +37,12 @@
     "@rise-tools/kit-tamagui-toast": "0.2.0",
     "@rise-tools/kit-lucide-icons": "0.2.0",
     "@rise-tools/kit-svg": "0.2.0",
-    "@rise-tools/kit-qrcode": "0.2.0",
-    "@rise-tools/kit-expo-router": "0.2.0"
+    "@rise-tools/kit-react-navigation": "0.2.0",
+    "@rise-tools/kit-qrcode": "0.2.0"
   },
   "peerDependencies": {
+    "@react-navigation/native": "^6.1.17",
```
ericvicenti, line 44: maybe should be an optional peerDependency

ericvicenti (approved):
exciting!

we should probably use react native screens for a more 'native' feeling experience

more important: we need to implement deep linking so people can still connect via QRCode

grabbou:
> we should probably use react native screens for a more 'native' feeling experience

We're using `react-navigation/native-stack`, so I guess that is as good as it gets? Deep Linking has been added in latest commit, we should probably add support for QR Codes later, will create a dedicated issue for it.

PR #138, opened 2024-06-24
feat: rise icon is a dropdown + qrcode sharing screen

grabbou:
- Adds Popover dropdown to Rise icon with some options
- Adds new "modal" screen to show QR Code and options to share

### DIFF
--- a/apps/mobile/src/screens/connection.tsx
+++ b/apps/mobile/src/screens/connection.tsx
@@ -15,7 +15,7 @@ import {
 import { Rise } from '@rise-tools/react'
 import React from 'react'
 
-import { BackButton } from '../back-button'
+import { BackButton, DismissButton } from '../back-button'
 import { Connection, useConnection } from '../connection'
 import { DataBoundary } from '../data-boundary'
 import { useModelSource } from '../model-sources'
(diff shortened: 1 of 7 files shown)


apps/mobile/src/screens/connection.tsx
```diff
@@ -48,27 +50,31 @@ export function ConnectionScreen({
...
         getId={({ params }) => params.path}
-        initialParams={{
-          id: route.params.id,
-        }}
         options={({ route }) => ({
           title: route.params.path,
         })}
-      />
+      >
+        {({ route }) => <RiseScreen connection={connection!} path={route.params.path} />}
```
grabbou, line 67: Did a small consulting session with Satya regarding TypeScript and the way we manage params and decided to change from "Component" and "initialParams" to [passing params explicitly as per the documentation](https://reactnavigation.org/docs/hello-react-navigation/#passing-additional-props). 

This makes our code much cleaner and allows us to write `navigate('qr-code')` from within BackButton without having to read and pass connectionId, that we already know anyway (in context of current navigation stack).

Not saying this is final pattern, but it's definitely much cleaner and gets the job done!

ericvicenti (approved):
🔥

grabbou:
Added "Edit" button to the dropdown for non-built in connections. Thanks for review. Merging this and continuing with the release.

PR #139, opened 2024-06-28
feat: support dark mode

grabbou:
- update React Navigation config
- set root view background color to match React Navigation header background color to ensure modal animations do not flicker contrasting app background
- inline appearance of controls example (padding, button group)

### DIFF
--- a/apps/mobile/src/index.tsx
+++ b/apps/mobile/src/index.tsx
@@ -1,13 +1,9 @@
-import {
-  DarkTheme,
-  DefaultTheme,
-  NavigationContainer,
-  ThemeProvider,
-} from '@react-navigation/native'
+import { DarkTheme, DefaultTheme, NavigationContainer } from '@react-navigation/native'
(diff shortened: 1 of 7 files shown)


apps/mobile/src/index.tsx
```diff
@@ -26,6 +22,11 @@ function App() {
   })
 
   const scheme = useColorScheme()
+  const theme = scheme === 'dark' ? DarkTheme : DefaultTheme
+  useEffect(() => {
+    SystemUI.setBackgroundColorAsync(theme.colors.card)
```
ericvicenti, line 27: oh tricky, I guess this is helpful when the screen rotates and you see the background temporarily?

ericvicenti (approved):
🚀

apps/mobile/src/index.tsx
grabbou: Yeah, I didn't think about it, but you are right - that's another case. In our case, popping modal view was showing white background!

PR #141, opened 2024-06-29
feat: add dropdown menu to the Home Screen

grabbou:
Required for Play Store submission to link to our Privacy Policy. Note we don't really collect anything, so this is just something I generated and it worked well for App Store.

We're using this service and their live hosted variant to make it a bit easier for us to navigate this otherwise confusing space 😆


ericvicenti (approved):
nice, thanks!

PR #143, opened 2024-06-30
feat: live code for the conference talk + some tweaks

grabbou:
This is what we will - more or less - achieve on stage. Also, tweaked the RadioGroup a bit to allow horizontal shape.


ericvicenti (approved):
great. as discussed we should break out the 1-5 into a new RatingField component

grabbou:
True. Moved it into as separate "server function".

PR #144, opened 2024-07-02
fix: android styling

grabbou:
- fix: missing adaptive icon (uses Expo default)
- fix: status bar not matching the header background
- fix: status bar not matching splash screen
- fix: header back button spacing (styling borrowed from react-navigation)


grabbou:
Screenshot 2024-07-02 at 09 25 09
Screenshot 2024-07-02 at 09 24 13

ericvicenti (approved):

PR #145, opened 2024-07-02
breaking: restructure event types

ericvicenti:


TheKohan:
Thanks for that and for the previous doc's. The naming is so much nicer, less brain processing needed to read that for sure 😄

grabbou (approved):
Love this! Great improvements. Response was never "ModelState" as much as e.g. Component or Ref, so this is very good change!

PR #154, opened 2024-07-13
feat: `@rise-tools/cli`

arpitBhalla:


ericvicenti:
This is super exciting, @arpitBhalla! What parts are ready for review vs currently a work in progress?

Also, how does this PR compare to the RFC here? https://github.com/rise-tools/rise-tools/discussions/153

Don't worry if its not perfect, I would rather quickly merge something that isn't 100% polished, than spend a long time in code review.

We are doing a talk at TheGeekConf on July 25. Hope we can get this ready in time for us to demo on stage!

arpitBhalla:
@ericvicenti It covers all the commands and options mentioned in RFC, here are few points I think it still misses
- Doesn't watch any file other than model.tsx
- Doesn't work with alias/references. 
	- Can have a `rise.config.ts` with these configs?
	- Read tsconfig directly? 
- Tunnel isn't there yet.

packages/cli/src/start/server.ts
```diff
@@ -0,0 +1,199 @@
...
+import { type AnyModels, handleGetRequest } from '@rise-tools/server'
+import chokidar from 'chokidar'
+import Fastify from 'fastify'
+import internalIp from 'internal-ip'
+import qrcode from 'qrcode-terminal'
+import { WebSocketServer } from 'ws'
+import z from 'zod'
+
+// TODO
+import { connectWebSocket, createWSServerContext } from '../../../server/src/ws-connection'
```
arpitBhalla, line 13: @ericvicenti can we export these from `@rise-tools/server`?

grabbou:
Hey @arpitBhalla! Thanks for the PR! I am going to continue with my review, but meanwhile, I am answering some of your questions with my thoughts. Hope this helps!

> It doesn't support aliases or references.

This would be supported out of the box with `tsconfig.json`, right? (more on that later)

> Can we include a rise.config.ts for these configurations?
> Can it read tsconfig directly?

I think it would be great if we followed what others do as well:

- You can use Rise CLI without TypeScript
- If you want TypeScript, create `tsconfig.json` in the top-level and import our preset
- Our preset configures defaults such as JSX support with our custom factory
- We read that config file and apply it to the current project

I am also in favour of having Rise config, most likely via `cosmiconfig` so that we support various formats, from `.riserc`, to `rise.config.js` ending up with `rise.config.ts` or `rise` entry in the `package.json`. I think there are numerous different preferences in the community, and following example of libraries such as ESLint, we can adopt the same.
 
- if `tsconfig.json` is present in the root directory of the project, use that configuration  

Having a top-level `tsconifg.json` is important for various tools to detect and support TypeScript properly. So, based on previous suggestions, I would make this a requirement. This is aligned with [how Expo does things](https://docs.expo.dev/guides/typescript/#add-base-configuration-with-tsconfigjson) and I am sure - other libraries too.

> Tunnel functionality isn't implemented yet.

For this one, I am going to write up another RFC in a second. We had a discussion with Eric how this could potentially work!

packages/cli/src/start/server.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    server.listen({ port }).then(() => {})
+    return {
+      close() {
+        server.close()
+      },
+    }
+  }
+
+  wsServer(port: number) {
```
grabbou, line 129: question: Why don't we just do:
```ts
return createWSServer({ models: this.modelManager.getModels(), port })
```
Was there any particular reason? Because if not - then, we don't need to import those "lower-level" functions from `ws-server`, but can use top-level export instead.

grabbou: Same goes for `httpServer` method. It looks very similar to `createHTTPServer` and I wonder if we could use that one directly instead.

packages/cli/src/start/server.ts
```diff
@@ -0,0 +1,199 @@
...
+      // Close the watcher if we are in prod mode after adding all models
+      fsWatcher.close()
+    }
+
+    this.startServer()
+
+    console.log('Server started on', `${host}:${port}`)
+    console.log('Scan the QR from rise playground')
+
+    this.createDevQR(`rise-playground://${host}:${port}`)
```
grabbou, line 87: nit: I am sure Eric will know better (my memory is fading away), but I think our link protocol is just `rise://` 🙏

packages/cli/src/start/server.ts
```diff
@@ -0,0 +1,199 @@
...
+interface IndexedModels {
+  [key: string]: any
+}
+
+export const devArgsSchema = z.object({
+  ws: z.boolean().optional().default(false),
+  prod: z.boolean().optional().default(false),
+  port: z.number(),
+  host: z.string(),
+  cwd: z.string().optional(),
```
grabbou, line 25: BTW, now that we have this nice Zod schema, it would be great (as a follow-up) to automatically validate passed options and throw an error if their values are missing. There's also nice feature of Zod `describe` which would allow us to attach description to each property.

example/demo-cli/navigate.d.ts
```diff
@@ -0,0 +1,8 @@
+import '@rise-tools/kit-react-navigation/server'
+
+declare module '@rise-tools/kit-react-navigation/server' {
+export interface NavigatePath {
+'':string,
+'feedback-form':string,
```
grabbou, line 6: Question: should this file be git-ignored?

grabbou (approved):
Great work! Thanks for this massive PR! I am leaving some comments, mostly nits and questions, to better understand the scope of this PR. Hope it's helpful! Available on Discord if you want to chat later as well 🤙

packages/cli/src/start/server.ts
grabbou: I have a question about this here: https://github.com/rise-tools/rise-tools/pull/154#discussion_r1677033356

example/demo-cli/models/_sections/Groceries.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+
+import { Content } from '../_components/Content'
+import { Icon } from '../_components/Icon'
+import { Section } from '../_components/Section'
+import { Title } from '../_components/Title'
```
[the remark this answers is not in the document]
ericvicenti: I think there is some issue with using typescript paths that @arpitBhalla mentioned. Not sure if/how that can be fixed

packages/cli/src/eject/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  const watcher = chokidar.watch(WATCH_PATH, {
+    ignored: IGNORED_PATH,
+  })
+
+  const importScripts: string[] = []
+  const modelObject: string[] = []
+  let types = 'export type Model ='
+
+  watcher.on('add', async (modelPath) => {
+    const modelName = modelPath.split('/').slice(1, -1).join(':')
```
ericvicenti, line 17: so, its possible to preserve the slash in the real models using nested objects.

if you have a models folder which contains

```
foo.tsx
bar/baz.tsx
```

your model definition file would be:

```tsx

const models = {
  foo,
  bar: { baz }
}
```

so it would be great to support that at some point

packages/cli/src/eject/index.ts
```diff
@@ -0,0 +1,39 @@
...
+    modelObject.push(`'${modelName}': ${modelVar}`)
+    types += `'${modelName}'|`
+  })
+
+  watcher.on('ready', () => {
+    console.log("Created 'server.ts' file")
+    fs.writeFileSync(
+      'server.ts',
+      importScripts.join('\n') +
+        '\n\n' +
```
ericvicenti, line 29: instead of formatting manually, we could run the file through prettier to make sure it looks good. this will be even more handy when the code generation gets more advanced.

packages/cli/src/start/server.ts
ericvicenti: yes the protocol is just `rise:` and also the URL format is significantly different. See the code here to understand how rise URLs are actually created: 

https://github.com/rise-tools/rise-tools/blob/main/apps/mobile/src/screens/qr-code.tsx#L19-L20

basically we take a "Connection" object with several fields, json stringify it, base 58 encode it, then pass as a query param to `rise://connect?connectInfo=...`

packages/cli/src/start/server.ts
ericvicenti: we also have `createServer()` which supports both HTTP and WS. Not sure why we wouldn't just always use that one

packages/cli/src/start/server.ts
```diff
@@ -0,0 +1,199 @@
...
+    const { key } = this.parseModelPath(modelPath)
+    this.models[key]
+  }
+
+  getModels() {
+    return this.models
+  }
+}
+
+export class DevServer {
```
ericvicenti, line 60: I'm not sure why we need to use a class for this. Because this is a CLI, there will only ever be one instance.

I would like to avoid the complexity of classes, if possible. All the properties here can just be module scoped variables.

Same for ModelManager.

If we do want to support multiple dev servers, we could just use a function. Something like this:

```ts
function createDevServer() {
  let internalState = ...
  function start() { ...  }
  return { start }
}
```

We can keep the classes for now if you feel strong about them. But if Mike or I wrote this code, we would just use module state or functions. (See how the rise servers just use functions rather than classes)

ericvicenti (commented):
Very exciting!

You don't have to address all of these concerns before merging. We can continue to iterate after the PR has been merged

packages/cli/src/start/server.ts
arpitBhalla: I tried with hono initially, it's lightweight.

packages/cli/src/eject/index.ts
```diff
@@ -0,0 +1,39 @@
+import fs from 'node:fs'
+
+import chokidar from 'chokidar'
+
+import { IGNORED_PATH, WATCH_PATH } from '../config/constants'
+
+export const ejectAction = async () => {
+  const watcher = chokidar.watch(WATCH_PATH, {
+    ignored: IGNORED_PATH,
+  })
```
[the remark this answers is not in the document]
arpitBhalla: Reused `chokidar` to reduce number of dependencies.

packages/cli/src/eject/index.ts
arpitBhalla: It's great idea

ericvicenti:
Hey I think this makes sense and it's a huge step forward! Have you tested the QRCode with the real app?

One design principle to keep in mind, we want the users server.ts file to be as simple as possible. Maybe you can find a way to do QRCode with even less code?

example/demo/src/index.ts
```diff
@@ -1,9 +1,35 @@
...
+  const host = await getHost(HostType.tunnel)
+
+  // await createWSServer(models, port)
+  await createHTTPServer(models, port)
+
+  const qr = await createConnectionQR({
+    host: `${host}:${port}`,
+    label: 'Rise Demo',
+    id: 'demo',
+    path: '/',
```
ericvicenti, line 23: I think the default path is empty string, not "/"

example/demo/src/index.ts
```diff
@@ -1,9 +1,35 @@
...
 
-createWSServer({ ...inventory, ...controls, ...delivery }, Number(process.env.PORT || '3005'))
+const models = { ...inventory, ...controls, ...delivery }
+
+async function init() {
+  const port = Number(process.env.PORT || '3005')
+  const host = await getHost(HostType.tunnel)
+
+  // await createWSServer(models, port)
+  await createHTTPServer(models, port)
```
ericvicenti, line 17: Can we use createServer here?

Also, maybe the server should accept an argument for host

packages/cli/src/qr/index.ts
```diff
@@ -0,0 +1,20 @@
...
+export type ConnectionPayload = {
+  id: string
+  label: string
+  host: string
+  path: string
+}
+
+export const createConnectionQR = (connection: ConnectionPayload) => {
+  return new Promise((resolve) => {
+    const connectInfo = bs58.encode(Buffer.from(JSON.stringify(connection)))
```
ericvicenti, line 14: I think users may want to see the URL string, in addition to the QRCode. So we probably want a different API to support that

ericvicenti (changes requested):

packages/cli/src/getHost.ts
```diff
@@ -0,0 +1,30 @@
+import internalIp from 'internal-ip'
+
+import { getTunnelAddress } from './tunnel'
+
+export enum HostType {
```
TheKohan, line 5: nit: i personally don't like `enums` 😄 What makes the difference between  enum implementation and this -
```ts
const HostTypes = {
  lan : 'lan',
  localhost : 'localhost',
  tunnel : 'tunnel',
} as const
```
Because i (on the surface) don't find any, would gladly hear any `pros` of enums :smile:

example/demo-cli/models/_sections/Groceries.tsx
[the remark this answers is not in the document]
arpitBhalla: I found a way to do this

example/demo/src/index.ts
grabbou: I am surprised, but also excited, that `http` server works! Last time I tried it was broken! 😆

grabbou:
Let me post a quick update on the state of this PR for everyone who is following along:

We had a discussion on Discord (BTW, join if you haven't already) with @arpitBhalla and @ericvicenti where we decided to keep the workflow simple for now and not introduce a new CLI.

Instead, since we're running on the server, we will encapsulate all development logic (such as printing QR code, setting up the tunnel, or asking questions - if any) into `createServer` API. If you use that, you opt-in for a bunch of DX improvements that come with Rise.

If you prefer to run closer to the metal, you can import any of the publicly available APIs (such as createWSServer, handleRequest or createQRCode) and build up your own experience that works for your organisation. 

We're waiting for the final update that brings the API closer to the above description and we're shipping this 🔥

grabbou (approved):
Thank you a lot for your contributions. I am going to approve and import this PR to `dev` branch. After doing few additional features, I will merge this into `main` and release (likely tomorrow).

Thank you once again for your time and effort! This is very much appreciated! 🔥

arpitBhalla:
@grabbou 
Check this PR for some additional tunnel features that I've added
https://github.com/rise-tools/rise-tools/pull/159

PR #155, opened 2024-07-13
feat: `create-rise`

arpitBhalla:
templates are empty


packages/create-rise/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "description": "",
+  "main": "index.js",
+  "bin": {
+    "create-rise": "dist/cli.js"
+  },
+  "scripts": {
+    "dev":"tsx src/cli.ts"
+  },
+  "author": "",
+  "license": "ISC",
```
grabbou, line 13: nit: We should make sure before we merge the PR that it's all MIT 🗡️

packages/create-rise/src/cli.ts
```diff
@@ -0,0 +1,132 @@
...
+  const { projectName, overwrite } = answers
+  const root = path.join(cwd, projectName)
+
+  if (overwrite === 'yes') {
+    emptyDir(root)
+  } else if (!fs.existsSync(root)) {
+    fs.mkdirSync(root, { recursive: true })
+  }
+
+  const renameFiles: Record<string, string | undefined> = {
```
grabbou, line 54: When the value can be undefined? What is the meaning then?

grabbou, line 54: When the value can be undefined? What is the meaning then?

packages/create-rise/src/cli.ts
```diff
@@ -0,0 +1,132 @@
...
+    write(file)
+  }
+
+  console.log('Done')
+
+  if (projectName !== '.') console.info('cd', projectName)
+  console.info('npm install')
+  console.info('npm start')
+
+  function write(file: string, content?: string) {
```
grabbou, line 78: If we need these helper functions, maybe better to have a separate file (with utils) or at least move them outside of prompt function? Probably cleaner to understand (as declarations are hoisted otherwise) and will make it explicit that all values go in as function arguments

packages/create-rise/src/cli.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  const pkg = JSON.parse(fs.readFileSync(path.join(templateDir, `package.json`), 'utf-8'))
+
+  pkg.name = targetDir === '.' ? path.basename(path.resolve()) : targetDir
+
+  write('package.json', JSON.stringify(pkg, null, 2) + '\n')
+
+  const files = fs.readdirSync(templateDir)
+  for (const file of files.filter((f) => f !== 'package.json')) {
+    write(file)
```
grabbou, line 69: Question regarding performance: is it okay to leave this synchronous, or would it be better if `write` was async (and all `fs` functions were awaited), and e.g. files were copied with `Promise.all([write(), write()]`?

packages/create-rise/src/cli.ts
```diff
@@ -0,0 +1,132 @@
...
+      return true
+    }
+  }
+
+  function emptyDir(dir: string) {
+    if (!fs.existsSync(dir)) {
+      return
+    }
+    for (const file of fs.readdirSync(dir)) {
+      if (file === '.git') {
```
grabbou, line 119: Question: What is the behavior of other CLI libraries, e.g. Expo or Create React App, when it comes to overwriting? Does it overwrite, or leave Git? I am genuinely curious - haven't thought about this before!

grabbou (commented):
Looking good! Excited to see this up and running. Left some preliminary feedback to help you move forward. I am sure Eric will have his thoughts regarding prompts and what we ask the users, but it's looking really good!

packages/create-rise/src/cli.ts
```diff
@@ -0,0 +1,132 @@
...
+import path from 'node:path'
+
+import inquirer from 'inquirer'
+
+const cwd = process.cwd()
+
+const defaultProjectName = 'rise-project'
+
+async function prompt() {
+  let targetDir = ''
```
grabbou, line 12: I know this is totally out of scope for this PR, but maybe we print beautiful Rise logo here (and in CLI start command), like React Native! 😆

CC: @ericvicenti

packages/create-rise/src/cli.ts
arpitBhalla: The small number of files to copy should not cause any performance issues.

btw, I've copied the utils from `create-vite`

packages/create-rise/src/cli.ts
arpitBhalla: Copied the utils from `create-vite` 😅

packages/create-rise/template-react-navigation/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "main": "index.js",
+  "scripts": {
+    "start": "rise start",
+    "eject": "rise eject"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@rise-tools/server": "latest",
+    "@rise-tools/kitchen-sink": "latest"
```
grabbou, line 14: this is a totally non-blocking nitpick that we can fix in a follow-up, but just wanted to point out that:
1. This example should be renamed to `template-basic` or...
2. We should add `kit-react-navigation` as a dependency
Reason being that React Navigation is not included in the `kitchen-sink` as we wanted to make it routing library agnostic.

packages/create-rise/template-react-navigation/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "template-react-navigation",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "start": "rise start",
+    "eject": "rise eject"
```
grabbou, line 8: We will have to remove this in favour of `tsx` since we don't really have CLI shipped yet. This is also non-blocking and I will just approve, we can handle that in a follow-up.

grabbou (approved):
Looking good. Thanks for the PR! I will now merge it, test locally a few times, and publish that new package.

grabbou (approved):
Thanks! 🔥

PR #159, opened 2024-07-16
feat: tunnel

arpitBhalla:


tunnel/server.ts
```diff
@@ -0,0 +1,87 @@
+import { serve } from '@hono/node-server'
+import { Hono } from 'hono'
```
TheKohan, line 2: I too like `hono` ! Nevertheless i don't know if we shouldn't use `fastify` like on the `ws` server, just for clarity. Wdyt @ericvicenti / @grabbou

TheKohan:
Great job @arpitBhalla, this CLI/Starter/tunnel stuff is shaping up pretty nicely !

tunnel/server.ts
grabbou: Yes, I haven't looked at the PR just yet (busy with other PRs), but generally speaking we should maintain whatever are our existing solutions, and if we change them - change everywhere for a good reason!

PR #160, opened 2024-07-17
feat: release `create-rise` app

grabbou:
This PR builds on the previous work by @arpitBhalla (#155)! Once again, thank you for working on this! 👏

While working on the release of `create-rise` to npm, I implemented a few additional features:

- Add `create-rise` to our build pipeline (in the `tsconfig.build.json`) so its source files are transpiled during our release process. Previously, we would ship an empty `dist` folder, unless build command was run explicitly in that project
- Rename `dist` to `lib` to unify the naming convention with all other packages
- Create a dedicated package for TypeScript preset, so that users don't have to use `jsxImportFactory` themselves. Also, document this package in the docs (very briefly, but still)
- Configure JSX factory inside the template to use Rise, so that we don’t have to import React in there
- Move templates from within `create-rise/src` to be a standalone `npm` package. That way we can run them independently, test them, make sure they have all required dependencies listed in the `package.json`. Now, if you want to develop the template, just start it, load it in Rise, and add necessary UI.
- Implement missing commands for installing packages etc. To do so, use `zx` from Google, that includes async/await support, extended filesystem that supports JSON as well as built-in spinners. Use other helpers such as `spinner`, `chalk` etc. from that package as well. That allowed to remove some of the helpers we created the other day.
- Switched from CommonJS to ES Modules inside `create-rise`. I tried both `zx` and `execa` and they do not support CommonJS anymore (I expect this to be common in the Node world), so I switched to new module format ahead of time.
- Remove `ignore` option when folder we want to create already exists. Require to be empty, to avoid undefined behavior. Also, do not have special treatment for folders with just “.git”. If there’s existing folder, stop (aligned with Expo)

There are still a few TBDs in the code, some of them should be addressed before this package gets published. However, reviews are welcome!

Here is our very rough, but working experience:

Screenshot_2024-07-17_at_08 51 20

Note:
- We are not going to support multiple package managers right now (#161)


grabbou:
Unfortunately, I can't request review from either @TheKohan or @arpitBhalla via Github UI, so you are more than welcome to have a look!

example/demo/tsconfig.json
```diff
@@ -1,9 +1,5 @@
 {
-	"extends": "../../tsconfig.json",
-	"compilerOptions": {
-		"jsx": "react-jsx",
-    "jsxImportSource": "@rise-tools/react"
-	},
+	"extends": "@rise-tools/preset-typescript",
```
TheKohan, line 2: Just out of curiosity, shouldn't it be `@rise-tools/preset-typescript/tsconfig`, or would it work anyway since there's only one file listed in `preset-typescript` package.json

packages/create-rise/package.json
```diff
@@ -1,17 +1,42 @@
...
+  "author": "Arpit Bhalla (https://github.com/arpitBhalla)",
+  "contributors": [
+    "Mike Grabowski <grabbou@gmail.com> (https://github.com/grabbou)"
+  ],
+  "bugs": {
+    "url": "https://github.com/rise-tools/rise-tools/issues"
+  },
+  "homepage": "https://github.com/rise-tools/rise-tools#readme",
   "bin": {
     "create-rise": "dist/cli.js"
```
arpitBhalla, line 32:
```suggestion
    "create-rise": "lib/cli.js"
```

grabbou: Ha! Great catch!

TheKohan (approved):
Looks great, love the new way of handling templates, wouldn't have figured out 😅

example/demo/tsconfig.json
grabbou: I initially thought the same, but after reading TypeScript documentation, it appears that if there's top-level tsconfig.json, it is automatically loaded when you refer to it this way.

grabbou:
> Looks great, love the new way of handling templates, wouldn't have figured out 😅

Ha, credits go to Expo team - I had a look at the way they do things and it's pretty incredible. In their case, it's even better (but an overkill for us right now).

They have separate repository with all the examples and when you start the CLI, they load all of them and list them dynamically for you to choose. So it's very easy for them to ship a new template to use.

example/demo/tsconfig.json
TheKohan: Yea, was curious becuse of the default `extend` in expo, but that's not a explicitly defined `tsconfig.json` 
```json
"extends": "expo/tsconfig.base"
```
Good to know !

TheKohan:
> > Looks great, love the new way of handling templates, wouldn't have figured out 😅
> 
> Ha, credits go to Expo team - I had a look at the way they do things and it's pretty incredible. In their case, it's even better (but an overkill for us right now).
> 
> They have separate repository with all the examples and when you start the CLI, they load all of them and list them dynamically for you to choose. So it's very easy for them to ship a new template to use.

Yea was thinking about dynamically loading all templates (without explicitly define all of the templates in `cli.ts`). Nice to keep in mind.

TheKohan (approved):

packages/create-rise/src/utils.ts
```diff
@@ -1,56 +1,36 @@
...
-      continue
-    }
-    fs.rmSync(path.resolve(dir, file), { recursive: true, force: true })
+export async function downloadAndExtractTemplate(root: string, packageName: string) {
+  const response = await spinner('Downloading template', async () => {
+    const tarball = (await $`npm view ${packageName} dist.tarball`).stdout
+    return fetch(tarball)
+  })
+  if (!response.ok || !response.body) {
+    throw new Error('Failed to fetch the code for example from ${tarball}.')
```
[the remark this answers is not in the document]
grabbou: thanks!

PR #164, opened 2024-07-17
feat: improve `create-rise` styling

TheKohan:
image

Another PR to visually please our beloved users!
* Added icons, used `chalk` to make `create-rise` "pop"
* Added option to skip installing packages (`astro` does this, thought it would be nice if we don't support multiple package managers for now)
* Replaced `inquirer` with `@inquirer/prompts` its successor (with a bit cleaner api and access to `theme`  type) 
* Themed inquirer prompts with `stylePrompt` fnc

Feel free to add your opinion's in this PR.

### DIFF
--- a/packages/create-rise/src/cli.ts
+++ b/packages/create-rise/src/cli.ts
@@ -6,14 +6,19 @@ import { fileURLToPath } from 'node:url'
 import dedent from 'dedent'
 import gradient from 'gradient-string'
 import inquirer from 'inquirer'
-import { $, cd, chalk, fs, spinner } from 'zx'
+import { $, cd, chalk, fs, minimist, spinner } from 'zx'
 
 import { downloadAndExtractTemplate, formatTargetDir, isNodeError } from './utils.js'
+
+type Args = {
+  verbose?: boolean
+}
 
 const WELCOME_TO_RISE_ASCII =
   "__          __  _                            _          _____  _            _ \r\n\\ \\        / / | |                          | |        |  __ \\(_)          | |\r\n \\ \\  /\\  / /__| | ___ ___  _ __ ___   ___  | |_ ___   | |__) |_ ___  ___  | |\r\n  \\ \\/  \\/ / _ \\ |/ __/ _ \\| '_ ` _ \\ / _ \\ | __/ _ \\  |  _  /| / __|/ _ \\ | |\r\n   \\  /\\  /  __/ | (_| (_) | | | | | |  __/ | || (_) | | | \\ \\| \\__ \\  __/ |_|\r\n    \\/  \\/ \\___|_|\\___\\___/|_| |_| |_|\\___|  \\__\\___/  |_|  \\_\\_|___/\\___| (_)"
 
 async function createRise() {
+  const { verbose } = minimist<Args>(process.argv.slice(2))
   const { projectName, installDeps } = await inquirer.prompt([
     {
       type: 'input',
@@ -98,7 +103,10 @@ async function createRise() {
 
     console.log(chalk.bold(`📦 Using ${chalk.bold.cyan('npm')} to install packages.`))
 
-    await spinner(chalk.bold(`🔧 Installing dependencies in ${projectName}`), () => $`npm install`)
+    await spinner(
+      chalk.bold(`🔧 Installing dependencies in ${projectName}`),
+      () => $({ quiet: !verbose })`npm install`
+    )
   }
 
   console.log(riseGradient(WELCOME_TO_RISE_ASCII))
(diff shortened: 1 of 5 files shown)


packages/create-rise/src/cli.ts
```diff
@@ -4,19 +4,29 @@ import path from 'node:path'
...
     {
       type: 'input',
       name: 'projectName',
       message: 'Project Name',
       default: 'rise-project',
     },
+    {
+      type: 'confirm',
+      name: 'installDeps',
+      message: 'Would you like to install dependencies ?',
```
grabbou, line 27: nit: space before ?

packages/create-rise/src/cli.ts
```diff
@@ -48,7 +58,10 @@ async function createRise() {
       },
     ])
     if (!overwrite) {
-      console.error(`The directory "${root}" already exists. Exiting...`)
+      console.error(dedent`
+        ${chalk.red.bold('🚨 There was an error setting up new Rise project:')}
+        ${chalk.red(`The directory "${root}" already exists. Exiting...`)}
+        `)
```
grabbou, line 64: nit: make sure to run ESLint on save. Line 64 looks like bad indentation which should've been fixed by ESLint already!

packages/create-rise/src/cli.ts
```diff
@@ -77,14 +90,29 @@ async function createRise() {
...
+    console.log(chalk.bold(`📦 Using ${chalk.bold.cyan('npm')} to install packages.`))
+
+    await spinner(chalk.bold(`🔧 Installing dependencies in ${projectName}`), () => $`npm install`)
+  }
+
+  console.log(riseGradient(WELCOME_TO_RISE_ASCII))
 
   console.log(
-    `The project has been successfully created in ${projectName}. To start, run 'npm dev'`
+    [
```
grabbou, line 107: nit: we can use `decent` and multiline string, e.g.
```
welcome

to

rise
```
instead of the array notation!

grabbou (approved):
Love it! Thanks for working on it!

To be honest, I would only add a configuration/theme where we keep primary/secondary colors and use them instead of default green/red for success/failure!

packages/create-rise/src/cli.ts
```diff
@@ -96,11 +124,13 @@ async function copyAdditionalTemplateFiles(root: string) {
 
 createRise().catch((e) => {
```
grabbou, line 125: Meanwhile @TheKohan while we wait for @ericvicenti thoughts on the logo, I have another quick one for you!

We should switch $`npm install` to be $({ quiet: !verbose })`npm install` and use `minimist` from `zx` to check for that flag and allow passing `--verbose`.

`npm install` generates excessive output and will make the beautiful UI of the CLI you created much uglier!

grabbou (approved):

arpitBhalla:
Can get pkg manager using `npm_execpath` env var

```ts
const packageManager = () => {
  const execPath = process.env.npm_execpath;

  if (execPath.includes('yarn')) {
    return 'yarn';
  } else if (execPath.includes('pnpm')) {
    return 'pnpm';
  } else {
    return 'npm';
  }
};
```

TheKohan:
> Can get pkg manager using `npm_execpath` env var
> 
> ```ts
> const packageManager = () => {
>   const execPath = process.env.npm_execpath;
> 
>   if (execPath.includes('yarn')) {
>     return 'yarn';
>   } else if (execPath.includes('pnpm')) {
>     return 'pnpm';
>   } else {
>     return 'npm';
>   }
> };
> ```

We wanted to wait and try to port that logic from `react-native-cli`, there's been some discussion going on discord

arpitBhalla:
Got it. I saw npm install in the code and thought we would use npm i until the PR is merged into that repo.

packages/create-rise/src/cli.ts
TheKohan: You meant `dedent` ? I've tried it but didn't know how to conditionally display `installDeps` without making empty line when the value is not there.

grabbou:
> Got it. I saw npm install in the code and thought we would use npm i until the PR is merged into that repo.

Yes, that is the idea. Until the PR in the React Native Community is created/merged, we're just using `npm`. We may consider adding `--no-install` flag at some point down the road, as a temporary solution, but I don't think it's a problem for now!

packages/create-rise/src/cli.ts
TheKohan: done

packages/create-rise/src/cli.ts
TheKohan: Weird, it didn't pick it up as an error. My `vscode` should pick the `.vscode/settings.json` by default i think 🤔

TheKohan:
@grabbou  @ericvicenti  I've updated the look 😄

PR #166, opened 2024-07-18
feat: small improvements to the CLI

grabbou:
- Added `--help` command to print out available options
- Moved `Skip installing dependencies` from question to a flag (`--no-install` - similar to what others are already doing)
- Updated color for "defaultAnswer" so that default answers are not coloured same way as your own
- Using dedent in few more places
- Use gradient for highlight color
- Do not print error messages when you force close / interrupt running script while prompt is asking you a question

### DIFF
--- a/packages/create-rise/src/cli.ts
+++ b/packages/create-rise/src/cli.ts
@@ -7,7 +7,7 @@ import { confirm, input } from '@inquirer/prompts'
 import dedent from 'dedent'
 import { $, cd, chalk, fs, minimist, spinner } from 'zx'
 
-import { bold, debug, error, highlight, prompt, riseGradient, text } from './theme.js'
+import { bold, debug, error, gradient, highlight, prompt, text } from './theme.js'
 import { downloadAndExtractTemplate, formatTargetDir, isNodeError } from './utils.js'
 
 type Options = {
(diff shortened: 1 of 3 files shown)


packages/create-rise/src/cli.ts
```diff
@@ -122,10 +114,39 @@ async function copyAdditionalTemplateFiles(root: string) {
   await $`cp ${path.join(source, '../_gitignore')} ${path.join(root, '.gitignore')}`
 }
 
-createRise().catch((e) => {
+const opts = minimist<Options>(process.argv.slice(2), {
```
arpitBhalla, line 117: Can add a `--non-interactive` flag, might helpful in CI.

https://docs.expo.dev/more/create-expo/#--yes

packages/create-rise/src/theme.ts
```diff
@@ -1,39 +1,23 @@
...
-      prefix: colors.base('>'),
-      style: {
-        answer: colors.highlight,
-        message: colors.base,
-        defaultAnswer: colors.highlight,
-        error: colors.error,
-      },
-    },
-  }
+export const prompt = {
```
TheKohan, line 16: WEEEELLLL i might have over-engineered things 😂

packages/create-rise/src/cli.ts
grabbou: feel free to send over in a separate PR!

TheKohan (approved):

PR #168, opened 2024-07-18
feat: various tweaks to the CLI package

grabbou:
This PR builds on top of recently merged CLI works by @arpitBhalla (once again, thank you for your contributions).

Here are a few additional features that I implemented on top, before merging to main:
- **Just like in the recent PR, I am migrating all our Node.js modules to use ESM.**
  - That way, we can import latest packages (e.g. 7.x of internal-ip, not 4.x) that are ESM-only.
  - ESM is compatible with Common.js, but not the other way around. 
  - It means that we can still import all our Common.js packages that target React/React Native without any issues. 
  - Going forward, we should most likely adopt ESModules for the entire project (for clarity) and transpile to CJS with Babel.
- Removed "moduleResolution" "bundler" in "server" package that I believe ended up there by a mistake (copy pasta 👀). We don't really use bundler here. In fact, it was making TypeScript output "export" syntax, without flagging package as "type: module". I removed this and it now provides CommonJS, which is aligned with every other non-Node package right now.
- Move `theme` from `create-rise` to `cli` package, so we can share things such as logo or styling across different packages
- Replace `createServer` API from the `cli` with `@rise-tools/server` package. Right now, we're using `createWSServer`, since `createServer` has some issues (due to http being broken). This is on hold.
- Replace dummy values for connection name and label with values from project's package.json
- Removed some extraneous dependencies from various packages

In a follow-up PR, I am going to apply those changes to http data source, provided it can be fixed on the spot. Then, we will move onto the tunnel work.

Screenshot 2024-07-18 at 23 21 07


grabbou:
TBD: Fix failing tests due to ESModules & Jest.

example/demo/src/index.ts
```diff
@@ -1,25 +1,13 @@
...
-}).then((server) => {
-  console.log(server.appConnectionURL)
-  server.printQR()
-})
-
-declare module '@rise-tools/kit-react-navigation/server' {
-  export interface Navigate {
-    path: keyof typeof models
-  }
-}
```
arpitBhalla, line 25: @grabbou Can we keep this? It doesn't work with lookup right now, but it'd help in typing `navigate()`. What do you think?

grabbou: Oh yes, I will look into this. We should have this in the demo as well as default template. Thanks for pointing this out!

PR #171, opened 2024-07-18
feat: migrate from jest to vitest

grabbou:
Reasoning:
- Poor support for ESModules (still marked as experimental)
- Using `ts-jest` makes things much slower
- Relying on Babel

Switching over to Vitest:
- Much faster and interactive
- Same API
- Easy and clean configuration for node and react packages (no more "jest-environment")
- Out of the box support for TypeScript, ESM and others (via ESBuild)
- Better support for multiple workspaces (and dedicated environments)

Other notable changes in this PR:
- Removed exclusion of `__tests__` from TypeScript project to provide better type safety (note: tests are not included in the npm package)
- Removed `node-fetch` in tests - it's built-in in all Node versions we support
- Cleaned-up Babel configs (we never needed them)


grabbou:
After spending exactly 24 hours on this, I am proud to call myself ESM/CJS interop master (just kidding - this will get back to us later anyway).

PR #175, opened 2024-07-19
feat: infer nested & lookup model keys

arpitBhalla:


grabbou:
Thanks for the PR!

> In future, can improve the types to support params

What do you mean by this? Isn't this PR supporting params by expanding lookup into `<modelName>/${string}/...` as many times as the lookup is nested?

arpitBhalla:
Yes, it does

I thought params would be an Object in future

grabbou:
Ah, no! That was just my moment of stupidness! Paths are way to go! :fire:

grabbou (approved):
This is simply AWESOME! I hate those generics, spent so much time working on them last time... thanks! 🔥

PR #176, opened 2024-07-19
feat: set navigation options ahead of transition 

grabbou:
Allows to set title and configure other screen options ahead of transitioning, which deals with flickering title etc. We will have to further improve this later to support other types of navigators, but this is good enough for now!

This PR also cleans-up two additional files that got into bad state after recent rebase/merge.


grabbou:
Merging and will soon follow up with #175 that improves type inference (I can remove hardcoded `/inventoryItems` thing)

PR #177, opened 2024-07-19
readme: added build step before running the playground/demo server.

bmanojsai:
Hey,
Noticed the demo server or the playground requires the builds for the packages. Adding the build command info to the Contributor doc's page. Starting small with this change and looking forward to working on bigger PRs later 😄. let me know if I need to change anything like the position or styling of the info.

### DIFF
--- a/docs/docs/contributors.md
+++ b/docs/docs/contributors.md
@@ -20,14 +20,6 @@ cd rise-tools
 npm install
 ```
 
-## Generate Build
-
-After installing the dependencies, you need to generate the build for the project if you want to run the playground or the demo server:
-
-```sh
-npm run build
-```
-
 ### Run Tests
 
 `npm run test`
@@ -47,7 +39,9 @@ For Android: `npm run android`
 
 ### Run Demo Server
 
-First `cd example/demo`
+First, make sure you are in root folder of the project and run `npm run watch` to start the build watcher to view your changes live.
+
+then `cd example/demo`
 
 `npm run dev`
 
--- a/docs/docs/contributors.md
+++ b/docs/docs/contributors.md
@@ -20,6 +20,14 @@ cd rise-tools
 npm install
 ```
 
+## Generate Build
+
+After installing the dependencies, you need to generate the build for the project if you want to run the playground or the demo server:
+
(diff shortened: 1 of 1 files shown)


grabbou:
Hey! Welcome and thank you for the first PR! It is indeed undocumented behavior and I am terribly sorry for not putting it myself in Contributing.md in first place.

The problem is that `tsx` (that we use to run the server) does not respect `tsconfig.json` "paths"... so it won't get the code from source, like React Native's Metro or Vitest.

I am thinking about improving this, but for now, as a good practice, it is best to run `npm run watch` in the background while you're developing. We may remove the paths behavior in the future, since it doesn't work that well with demo server!

docs/docs/contributors.md
```diff
@@ -20,6 +20,14 @@ cd rise-tools
 npm install
 ```
 
+## Generate Build
+
+After installing the dependencies, you need to generate the build for the project if you want to run the playground or the demo server:
+
+```sh
+npm run build
+```
```
grabbou, line 29: Would be great to mention that its best to run `npm run watch` afterwards to continue seeing changes while working on server-related things.

Here's some context:
- We generally use `tsconfig.json` paths to rewrite paths from `lib` to `src`. That way, in development, we can simply load source code without having to start compilation. If you're running the React Native application (Playground) or running tests with Vitest, they both get raw TypeScript and compile it on the go.
- Unfortunately, `tsx` (that we use to run the demo server) does not recognise paths, and as a result, will always fallback to loading `lib`. 

TL;DR 

Easy version: **always keep npm run watch in the background**

grabbou (approved):

grabbou:
PS. Please join us on Discord! https://discord.gg/gdwtbJy7gp

docs/docs/contributors.md
bmanojsai: got it. 

changed the build command to watch and moved the info into the 'Run Demo Server section' as it's required only there. I will see why the tsx is not working even though its explicitly provied as `tsx watch ./src`.

grabbou:
Thank you!

PR #178, opened 2024-07-19
feat: infer lookup path for nested models and add unit tests

grabbou:
- Added tests for types (OMG, I love `vitest` so much!)
- Added support for nested/nested lookup models (wasn't working for some reason)
- Removed support for "objects" of models (this is currently not supported, we can always add it back)

CC: @arpitBhalla

### DIFF
--- a/packages/server/src/types.ts
+++ b/packages/server/src/types.ts
@@ -34,26 +34,19 @@ export type AnyModels =
   | ValueModel<any>
   | LookupModel<any>
 
-type LookupModelPathOf<M, Prefix extends string = ''> =
-  M extends ValueModel<any>
-    ? `${Prefix}/${string}`
-    : M extends LookupModel<infer LM>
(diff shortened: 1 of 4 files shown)


packages/server/src/types.ts
```diff
@@ -34,26 +34,19 @@ export type AnyModels =
...
   | LookupModel<any>
 
-type LookupModelPathOf<M, Prefix extends string = ''> =
-  M extends ValueModel<any>
-    ? `${Prefix}/${string}`
-    : M extends LookupModel<infer LM>
-      ? LookupModelPathOf<LM, `${Prefix}`>
-      : M extends object
-        ? InferModel<M, `${Prefix}/${string}/`>
-        : never
```
arpitBhalla, line 44: An object case is necessary if the lookup returns an object. @grabbou

PR #173, opened 2024-07-19
feat: tunnel 

arpitBhalla:

PR #179, opened 2024-07-20
umbrella: cloud / tunnel

grabbou:
Check each commit for details:
- #173


grabbou:
Let's keep this one on hold for now, as we're going to go with smaller implementation for now - as per #181. We still want to implement authentication as per original write-up, but we need a bit more time and thinking into this.

grabbou:
Going to close this as we ended up using hashes in our proxy service instead of this. While this is super cool, we prefer the slimmer alternative we have right now - less services to maintain, more stability!

PR #180, opened 2024-07-20
revert: migrate node packages to be ESM only

grabbou:
Initially, after running into compatibility issues using Sindre modules, I was under impression that ESM was the golden standard.

Turns out - it's not, and the last package release to support CJS is actually the one making 80% of total downloads, which means there's huge community that needs CJS. 

Screenshot 2024-07-20 at 14 12 44
Screenshot 2024-07-20 at 14 12 31

[Here is some additional context](https://www.reddit.com/r/node/comments/14rg9ym/esm_not_gaining_traction_in_backend_node/)

Since ESM is not compatible with CJS (but CJS is with ESM), it is better to pin projects that caused us to refactor in first place. We can revisit this later.

Note: We could also revert commit that migrates from Jest to Vitest, since we don't really need it now. However, Vitest is faster and supports testing types and has Jest-compatible API, so I am not going to do it for now.


grabbou:
Note: `zx` is set as `type: module` but does export CommonJS files, that's why it works with `@ts-ignore`. There is huge fragmentation in the ecosystem as to how to handle dual packaging.

PR #181, opened 2024-07-21
feat: automatically expose connection via Rise Proxy

grabbou:
Given time constraints, this is much simpler and smaller implementation of original Tunnel idea we had in mind.

Key highlights:
- DigitalOcean VPS running [Sish](https://github.com/antoniomika/sish) via Docker Compose (I took me a while to figure this out). We're on $6 plan a month. We will see how long it lasts.
- We use [SSH tunneling](https://www.ssh.com/academy/ssh/tunneling)
- We use `*.proxy.rise.tools` subdomain

Benefits:
- No additional tools required. Just SSH client.
- No cloud service involved. Less things to maintain. We will eventually need it (more on it later).

How this works:
- Project is assigned its unique ID (via `uuid.v4()`) 
- Project ID is stored in `.rise/projectId` file
- Tunnel will be open on each server start automatically, with 5s timeout 
- Hostname is fixed to be <projectId>.proxy.rise.tools.

Known limitations:
- URL must be unique, so we use rather ugly projectId.proxy.rise.tools. It's not that big of a deal, since you scan QR Code or copy the deep link
- Anyone can start the tunnel knowing your projectId. It means that it's not 100% secure (although brute force guessing your projectId is only theoretically possible). However, if you share your connection with network, it may become more likely. **That's why tunnelling should only be used for development**

[All these two limitations can be fixed easily](https://github.com/antoniomika/sish/issues/313) - I opened up an issue in the project we're using.

Once this is implemented, we will be able to:
- use nicer names for projects, e.g. "fancy-project"
- allow users to set custom names

**Preview:**
Screenshot 2024-07-21 at 02 03 32

TBD:
- Better UX if tunnel fails and while tunnel is loading (and it's been more than 2s already)
- Warn if no `''` model is present. QR Code points to index by default and this fails otherwise. This should be configurable.
- Provide default screen for example. When users add it, they will loose default Rise Playground screen with list of available examples. We should print it there instead.
- ONE DAY: maybe that default Playground screen can be Rise too!
- SSL

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,6 @@ lefthook.yml
 
 # Vitest
 **/*.vitest-temp.json
+
+# Rise
+**/.rise
--- a/package-lock.json
+++ b/package-lock.json
@@ -9089,6 +9089,11 @@
         "win32"
       ]
     },
+    "node_modules/@sec-ant/readable-stream": {
+      "version": "0.4.1",
+      "resolved": "https://registry.npmjs.org/@sec-ant/readable-stream/-/readable-stream-0.4.1.tgz",
+      "integrity": "sha512-831qok9r2t8AlxLko40y2ebgSDhenenCatLVeW/uBtnHPyhHOvG0C7TvfgecV+wHzIm5KUICgzmVpWS+IMEAeg=="
+    },
     "node_modules/@segment/loosely-validate-event": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/@segment/loosely-validate-event/-/loosely-validate-event-2.0.0.tgz",
@@ -9155,6 +9160,17 @@
       },
       "funding": {
         "url": "https://github.com/sindresorhus/is?sponsor=1"
+      }
+    },
+    "node_modules/@sindresorhus/merge-streams": {
+      "version": "4.0.0",
+      "resolved": "https://registry.npmjs.org/@sindresorhus/merge-streams/-/merge-streams-4.0.0.tgz",
+      "integrity": "sha512-tlqY9xq5ukxTUZBmoOp+m61cqwQD5pHJtFY3Mn8CA8ps6yghLH/Hw8UPdqg4OLmFW3IFlcXnQNmo/dh8HzXYIQ==",
+      "engines": {
+        "node": ">=18"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/sindresorhus"
       }
     },
     "node_modules/@sinonjs/commons": {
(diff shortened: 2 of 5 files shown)


arpitBhalla:
Does watching files attempt to reconnect the proxy?

grabbou:
> Does watching files attempt to reconnect the proxy?

Restarting server will simply restart the session.

We should likely decouple it, so we don't connect every time you save things. But this may require a bit more clever design - open to suggestions how this can be solved

arpitBhalla:
how about we create a `.pid` file to store the SSH process ID and run SSH in the background? On restart, we can check if SSH is still running with that process ID. If not, we can create a new SSH process.

grabbou:
> how about we create a `.pid` file to store the SSH process ID and run SSH in the background `{detached: true}`? On restart, we can check if SSH is still running with that process ID. If not, we can create a new SSH process.

Happy to accept a follow-up PR that does just that. I think that would be useful. No need to pollute our server with multiple SSH requests if we're reloading often.

Question: how would we clean this tunnel automatically if our server crashed and never restarted? Any way to make it close automatically?

arpitBhalla:
I think the options `ServerAliveInterval`, `ServerAliveCountMax`, and `ExitOnForwardFailure` might work. What do you think?

grabbou:
I ended up coming with much better, cloud-less solution. On new connection, on the server, when you create tunnel URL, do this:
- take requested user subdomain
- take user public fingerprint
- hash them

this is your URL. Return this URL to the client and use as a tunnel

Code:
```
hash := md5.New()
hash.Write([]byte(addr + sshConn.SSHConn.Permissions.Extensions["pubKeyFingerprint"]))
hashedAddr := fmt.Sprintf("%x", hash.Sum(nil))
proposedHost = fmt.Sprintf("%s%s.%s", hashedAddr, hostExtension, viper.GetString("domain")) 
```

This PR will remain unaffected. All changes will be done on the server.

grabbou:
Okay, I managed to modify the source of the proxy so that it hashes:

Log from the server
> Trying access: 0ba3656-dd69-40ae-b64d-49b2b8e3dd54, from fingerprint: SHA256:wCM0tvO/WJg+OkfaayQAKFAn7yT6Viw2w3LiG82QzWg, hash: f05ec229d85d7bafd545f0bbaeafffc9

The URL:

http://f05ec229d85d7bafd545f0bbaeafffc9.proxy.rise.tools/ 

URL is a hash of authorised public key fingerprint and passed projectId. 

TL;DR

only users that have your private/public key will be able to overwrite tunnel URL, which is expected.

arpitBhalla:
I absolutely loved how it turned out. You are a genius!

grabbou:
that's what team work is! discussions make it

grabbou:
In a follow-up, we need to change the behavior when connection fails.

Instead of throwing an error and quitting the process (after timeout of 5s), we should start normal server (without tunnel) instead. It might be likely that our tunnel goes down and we don't want everyone to not be able to develop with Rise. 

We should also print out an error (in grey color) that there was an error connecting and remote preview is not available.

We should also think about flag to disable/enable this behavior programatically.

PR #186, opened 2024-07-21
feat: add default screen for the example

grabbou:
Pretty much similar to our home screen

### DIFF
--- a/example/demo/src/home/ui.tsx
+++ b/example/demo/src/home/ui.tsx
@@ -1,0 +1,30 @@
+import { navigate } from '@rise-tools/kit-react-navigation/server'
+import { Button, YStack } from '@rise-tools/kitchen-sink/server'
+
+export const models = {
+  '': Home,
+}
+
+function Home() {
+  return (
+    <YStack padding="$4">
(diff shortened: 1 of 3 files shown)


example/demo/src/home/ui.tsx
```diff
@@ -0,0 +1,30 @@
+import { navigate } from '@rise-tools/kit-react-navigation/server'
+import { Button, YStack } from '@rise-tools/kitchen-sink/server'
+
+export const models = {
+  '': Home,
```
ericvicenti, line 5: this feels strange. instead I think we should `export function Home` and assign the empty string to Home inside of `src/index`

ericvicenti (approved):
Nice! Great to have this

Please change the import thing before merging.

ericvicenti:
Ok as we discussed, please add the <Rise> to show the list of examples

Then once you have a custom connection, have a greyed out hardcoded item that links to the new full screen home for the example server

grabbou:
I am going to +1 this comment today to have evidence tomorrow I knew what we meant 😆

PR #188, opened 2024-07-21
feat: provide info

grabbou:
You will see the following message in your terminal in case you didn't provide default path, or we were not able to pick model for the deep link.

In case `''` is missing, we will take one of existing models to start with that one.

Screenshot 2024-07-22 at 01 18 22


ericvicenti (approved):
Nice!

PR #189, opened 2024-07-23
fix(CLI): tunnel restart in watch mode

arpitBhalla:
- If master process is unchanged (tsx/nodemon) do no print instructions/restart tunnel. 
- added `--clear-screen=false` to prevent cleaning terminal in watch mode.


example/demo/src/index.ts
```diff
@@ -4,14 +4,15 @@ import { createWSServer, InferModel } from '@rise-tools/server'
...
+import { isProcessRunning } from './util'
 
 const models = { ...inventory, ...controls, ...delivery }
 
 const port = Number(process.env.PORT || '3005')
 
 const server = createWSServer(models, port)
 
-if (process.env.NODE_ENV === 'development') {
+if (process.env.NODE_ENV === 'development' && !isProcessRunning()) {
```
grabbou, line 15: I think a cleaner way would be to do this within `startTunnel`. If process is running, just print the URL, otherwise start it. I think end users shouldn't be aware of this, esp. in their entry file!

grabbou: That makes sense! Please apply that change!

PR #191, opened 2024-07-24
feat: replace hardcoded list of examples with partial Rise screen

grabbou:
- Do not hardcode examples, instead, use default screen for our example and render it in a partial view
  - To do so, we need to slightly overwrite "navigate" action in that partial screen to properly navigate into Rise navigator
- Remove "Share" option for default example. We always render it. Also, it would not work, because there's special handling for "example" ID anyways
- That said, dropdown now has just one option for builtin connection - "Go back". Decided to make the Rise logo always go back for builtin connection (no dropdown). Saves one extra click.
- When you have custom connection, the "Example" will render as always, at the end.

Questions:
- How should the "Example" be named? Right now, it's just "Example".

Screenshots:

### DIFF
--- a/apps/mobile/src/back-button.tsx
+++ b/apps/mobile/src/back-button.tsx
@@ -4,7 +4,7 @@ import { X } from '@tamagui/lucide-icons'
 import { Platform } from 'react-native'
 import { Button, ButtonProps, Image, Separator } from 'tamagui'
 
-import { Connection, EXAMPLE_CONNECTION } from './connection'
+import { Connection, DEMO_CONNECTION } from './connection'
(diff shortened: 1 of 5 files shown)


ericvicenti (approved):

ericvicenti:
Maybe "Example" should say "Rise Example UI" without the emoji, or if we want to have an icon it should be left-aligned in the button

grabbou:
I think icon-less is going to be easier for now. I don't think it's worth the effort to check if label contains emoji and add padding otherwise. Most of the connections won't really have an emoji in the label, maybe we can have a special field "icon" in the future!

Issue #170, opened 2024-07-18
feat(playground): splash screen improvements


TheKohan:
I've noticed that the splash screen on `playground` app is disappearing really fast on initial load, almost flickers.
Would be nice to extend the splash screen duration to like 1-2secs.
Some fancy splash animation would also be nice.

grabbou:
Feel free to send a PR! I would be happy to help!

TheKohan:
Most likely will pick that up later, just wanted to keep it here in backlog!

PR #196, opened 2024-07-27
feat(playground): added custom `Splash screen`

TheKohan:
As a followup to #170 i took a while to create this PR, let me know what you think.

* Removed image splash screen (left only bg colour)
* Added custom `splashscreen` implementation in JS with some animations.

Let me know what do you think, ofc all of the values timings are up to discussion 😄 

## With spin
https://github.com/user-attachments/assets/d7378289-99b0-4e54-9dd7-108a75559dc0

## Without spin
https://github.com/user-attachments/assets/c10088ba-b01b-4c3e-9f3d-56d0974e4bdd

### DIFF
--- a/apps/mobile/app.json
+++ b/apps/mobile/app.json
@@ -8,8 +8,6 @@
     "icon": "./assets/RiseIcon.png",
     "userInterfaceStyle": "automatic",
     "splash": {
-      "image": "./assets/RiseSplash.png",
-      "resizeMode": "contain",
       "backgroundColor": "#260162"
     },
     "updates": {
--- a/apps/mobile/src/index.tsx
+++ b/apps/mobile/src/index.tsx
@@ -9,6 +9,7 @@ import { GestureHandlerRootView } from 'react-native-gesture-handler'
 import { SafeAreaProvider } from 'react-native-safe-area-context'
 
 import { Screens } from './screens'
+import { SplashScreen } from './splash-screen'
 import { storage } from './storage'
 import { tamaguiConfig } from './tamagui/config'
 import { TamaguiProvider } from './tamagui/provider'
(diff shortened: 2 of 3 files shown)


arpitBhalla:
This looks awesome! How about sliding the logo upwards (along with rotation) to align it with the logo on the connection screen? What do you think?

TheKohan:
@arpitBhalla, I Was thinking about it, would for sure need some clever styling to stay aligned (and kept in mind doing Homescreen changes). Will try this out later today and post results here!

ericvicenti:
Cool! I would like to allow the app to open quickly, so if we do this I think it should be a very fast animation.

It would be cool if the splash screen would fade out so the UI doesn't flash and feel glitchy

Note, the connection screen will not be the default, if another connection was open last time the app was used.

TheKohan:
@grabbou  @ericvicenti 

I've speed up things (1 sec until content is visible), ignore metro loading. I'm not aligning the rise logo with LP @arpitBhalla, because we don't always open on LP -> 
image
Navigation state is persistent
https://github.com/user-attachments/assets/22c23fb7-9d9b-450f-9780-3df9ed87349b

TheKohan:
I've noticed a few issues with the code so i'll fix them later today, let me know in general if we want it or just drop it completely.

grabbou:
One comment:

I generally prefer when there's something visible on the Splash Screen immediately, to make it feel the app is snappier. As it's currently made, the logo will only start animating after JS finishes loading.

One way to workaround this would be to start with Rise logo visible by default (as it is right now on main) and start animating it after JS is ready. We could make it smaller/bigger (loop) and fade it all when we're ready.

Curious what you all think!

TheKohan:
I think i agree with @grabbou, will play around with i a bit. Nevertheless if we don't feel like doing fancy animations i can just revert to simple "wait until atleast fonts are loaded" scenario (previously splash wasn't waiting for anything).

grabbou:
> (previously splash wasn't waiting for anything).

Ha, interesting! Well, let's consider this bare minimum! I would double check with Eric as he has visual eye here - maybe on Discord!

grabbou:
Hey @TheKohan,

Do you have any updates on the PR?

PR #197, opened 2024-07-31
feat: easier deployment of dev server to production

grabbou:
While our demo server is not really production ready (yet), nor its this pipeline intended to be production ready, we did deploy our demo server to Render with tunnel option, so that everyone can continue accessing our demo from the conference by the QR code.

Unfortunately, that means [I had to create a separate Github repository, commit `.rise` folder to it and deploy the entire project.](https://github.com/rise-tools/vibes-example-server/tree/main/.rise) Otherwise, `setupRiseTools` wouldn't know what projectKey to use.

This PR adds ability to pass it via argument, which can e.g. be an environmental variable. That way, we can setup deployments without leaking secrets.

Once merged and released, I will open source (make public) repository with "vibes" example that is currently private due to key being committed.

I figured this might be useful for some people in the future too!

### DIFF
--- a/example/demo/src/index.ts
+++ b/example/demo/src/index.ts
@@ -13,7 +13,7 @@ const port = Number(process.env.PORT || '3005')
 const server = createWSServer(models, port)
 
 if (process.env.NODE_ENV === 'development') {
-  setupRiseTools({ server })
+  setupRiseTools({ server, projectKey: process.env.RISE_PROJECT_KEY })
 }
 
 import '@rise-tools/kit-react-navigation/server'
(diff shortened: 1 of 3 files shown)


ericvicenti (approved):
this looks good but didn't we decide to put the vibes example in with the other normal examples like inventory, etc?

grabbou:
Yeah, I will do that anyway! However, for the Vibes example itself, we need another entry-point that exports "vibes" as "home" (that's what the QR code in our Berlin talk has inside as "path"). 

Once we add "vibes" to full-size example, it will be under "vibes" and there will be different UI under "home" path. 

So, TL;DR:

- I will merge examples/ with templates/ and make it available in the CLI
- I will add vibes to example
- I may (although we will see in practice) create a separate project that imports "vibes" models and sets up new server with projectKey passed as argument, and "vibes" as "home" (this is what QR code from Berlin talk is). We should probably give this project explicit name, such as "geekconf-demo-server"

PR #198, opened 2024-07-31
feat: turn `example` into something people can start with

grabbou:
This PR makes it possible to get started with the full example as well.

I have removed simple-dev-server (counter example). Shall we merge the counter example with the rest, that is, create a separate folder for it and list it from `home`? I think it might not be necessary, given that we're adding "vibes" in next PR, that already demonstrate state.

Alternatively, we could put Counter in UI Controls.

As a follow-up, I will make it so that you can choose features from the full example, in the following manner:

- Get list of folders in `template/full/src` and treat each one of them as a separate part of example
- List them all in the CLI together with explanation what they do
- Let users choose
- Remove folders that were not selected by the user
- Remove imports that are not needed in the server entry point


grabbou:
Note: This will only work once we publish new example to `npm`.

ericvicenti:
Kind of a pain that the templates need to be individually published to npm. Can't we just grab the template code from this repo?

> I have removed simple-dev-server (counter example). Shall we merge the counter example with the rest, that is, create a separate folder for it and list it from home? I think it might not be necessary, given that we're adding "vibes" in next PR, that already demonstrate state.

Hm yeah, lets leave out the counter example because its so simple it is kind of boring

If we have several examples which showcase state, that is not a problem but I think they should be more interesting

ericvicenti (approved):
Ok yeah the publish to npm workflow fits well into our existing tooling. And it has the advantage of following the release cycle

PR #199, opened 2024-07-31
fix: events not working for '' path 

grabbou:
Javascript 😆 `if(!storeName)` will be true when `storeName` is an empty string, which makes events not run when within `''` model. 

Second check is enough. Anything that is not a string is invalid here, and `typeof` will handle both undefined/nulls just fine, in case they ever happen.

In fact, this check is necessary because we validate `event` with Zod and path will always be defined, otherwise it will throw an error elsewhere.


ericvicenti:
Ooops I wonder who fucked this one up! 🤣

PR #200, opened 2024-07-31
feat: improve create-rise DX to allow initing with a selected example

grabbou:
1. If you choose blank, you will start with template-blank.
2. If you choose full, you will start with template-full and all examples we have, including Home page that lists them all.
3. If you choose custom, you will have a select prompt to choose just one you get started with.

If you choose 2 or 3, we will use `template-full`.

In case 3, we will remove all examples but the one you selected, and we will update the imports in the `server.ts` to reflect that. 

To correctly locate imports and replace them, I annotated them with an inline comment. That inline comment gets removed and does not leak to the user space. 

Maybe there's better way, open to suggestions! 
Screenshot 2024-08-01 at 00 55 05
Screenshot 2024-08-01 at 00 55 19


packages/create-rise/src/cli.ts
```diff
@@ -73,9 +62,62 @@ async function createRise(opts: Options) {
...
+      'home',
+      ...AVAILABLE_EXAMPLE_CHOICES.map((e) => e.value).filter((name) => name !== example),
+    ]
+    for (const folder of toRemove) {
+      await fs.rm(path.join(templateRoot, folder), { force: true, recursive: true })
+    }
+
+    await fs.writeFile(
+      entryPointPath,
+      entryPoint.replace(MODELS_IMPORT_REGEX, `import { models } from './${example}/ui'`)
```
ericvicenti, line 112: Lol this is so hacky, but I'm not sure what else I would do to keep things somewhat clean

One obvious issue is that we should assign the appropriate example as the entry path `""`

Maybe we can find a cleaner way to do this, I'll think about it but lets chat more on discord

grabbou: Yeah, this is good old technique that we used back when we had first templates in React Native. It works and does not require any external library. This is also rather isolated use case and unless we plan to have more of this in the future, I think it's fine to have it done this way!

ericvicenti (approved):
Ok lets move forward with this approach because its pretty straightforward and doesn't have external dependencies

One other thing, if the user selects a single example, we should not copy the code for the other examples (or delete the other examples from the target project)

grabbou:
Yes @ericvicenti we do that here: https://github.com/rise-tools/rise-tools/pull/200/files#diff-fc477b2f72645be2695f01985dc26bdeffc29e8ccf92223e7f4a019ce6291cf2R102-R108

PR #202, opened 2024-08-02
fix: query model should auto-load on subscribe

ericvicenti:
I noticed the query model was not loading as expected when a client subscribes to it (or through a view)

This fixes the issue and includes a test to verify the behavior and prevent regressions


TheKohan (approved):

grabbou (approved):
LGTM :fire: So, we never loaded the query model, unless you called `invalidate`. This PR makes is so that `load` is called lazily on first subscribe. 

👍

PR #204, opened 2024-08-06
fix: get started link style

adham618:
Server-Defined-Rendering-for-React-Native-Rise-Tools

### DIFF
--- a/docs/src/pages/index.tsx
+++ b/docs/src/pages/index.tsx
@@ -36,7 +36,7 @@ function HeroSection() {
 
           <div className="flex items-center gap-4 max-lg:justify-center ">
             <Link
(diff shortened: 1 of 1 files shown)


ericvicenti:
Thanks for the PR! Guess we didn't test this on firefox 😅

In the future it would be helpful to explain what browser you found the issue on.

adham618:
@ericvicenti Chrome macbook

ericvicenti:
Oh, huh! For me it looks fine right now on Safari and Chrome, but on Firefox the button is not aligned

adham618:
This issue only occurs in older versions of browsers. After updating Chrome the problem disappeared

PR #209, opened 2024-08-13
fix: server component JSX allows returning null

ericvicenti:


grabbou:
Closing in favour of #212 that handles this case and a few others!

PR #210, opened 2024-08-13
feat: view model compare

ericvicenti:


grabbou:
My understanding is that this is a draft, but let me know what are the next steps here!

ericvicenti:
Ok this is good to go!

packages/server/src/model-view.ts
```diff
@@ -30,8 +31,18 @@ export function view<T>(
       return getModelState(model)
     }
     const newState = loadInput(getter)
-    state = newState
     isValid = true
+    if (opts?.compare) {
+      if (opts.compare === true && newState === state) return newState
+      if (
+        typeof opts.compare === 'function' &&
+        state !== undefined &&
```
grabbou, line 39: What is `state` is null? Is this possible? Do we want to run comparison function in that case? 
Also, is it possible that both `state` and `newState` are undefined? What should happen in such case?

packages/server/src/model-view.ts
```diff
@@ -30,8 +31,18 @@ export function view<T>(
       return getModelState(model)
     }
     const newState = loadInput(getter)
-    state = newState
     isValid = true
+    if (opts?.compare) {
```
grabbou, line 35: Should we provide a default comparator function as a very basic optimisation, just like React does with diffing? For example, `opts.compare` is always optional and can be a function, if not provided, we do simple equality check?

If we don't want this, what is the default behavior for users and question, whether users are not going to end up defining this themselves all over the place?

grabbou (approved):
Looking good. I have some inline questions as I am still expanding my understanding of that part of the codebase.

packages/server/src/model-view.ts
ericvicenti: state may be undefined if the previous state was empty because the dependencies were still loading. So probably the view is getting re-evaluated with the real state at this point

packages/server/src/model-view.ts
ericvicenti: yes if `opts.compare` is `true` we check with reference equality, which is fine for _some_ views

the default behavior is to always emit the new value to the subscribers, unless compare is the comparison fn or is true for cheking reference equality

Eventually, we will probably want a higher-level `view` that does deep comparison, which might be called `comparedView` or something, which will be handy for component views that you don't want to re-send to the client if it doesn't change

grabbou: I guess that's when we introduce `memo` type of view! Shall we document this in some advanced section of the documentation?

ericvicenti: Yeah as a follow up we can do that. Before I was trying to use `lodash.isEqual` for the deep comparison, but it wasn't working for some reason. So in my app I have verified this memoization is helpful by using `JSON.stringify(a) === JSON.stringify(b)` which is obviously bad, but I don't know what was wrong with isEqual.

PR #211, opened 2024-08-14
fix: streams: avoid extra re-subscription

ericvicenti:
This lack of memoization was causing aggressive re-subscription internally to rise


grabbou:
Yeah, note from the docs:

> If a different subscribe function is passed during a re-render, React will re-subscribe to the store using the newly passed subscribe function. You can prevent this by declaring subscribe outside the component.

TIL.

Thanks for spotting this!

PR #212, opened 2024-08-14
feat: allow all types of ReactNode JSX to be rendered on the server

grabbou:
You can now return more than just component definitions from within your server side helpers.


ericvicenti (approved):
Ok! We still don't support rendering Sets, but this is a good first step

grabbou:
Yeah, I will actually do it in a follow-up just in case, because otherwise I am afraid we will simply loose track of it! Little issues like this can be really annoying!

PR #213, opened 2024-08-14
feat: improves DraggableFlatList

ericvicenti:
- Use TouchableWithoutFeedback to mitigate the issue where the pressed state is stuck
- onPress callback for each item
- onPress callback for the whole component

Note: The issue is still present where the pressed state gets stuck down after dragging an item. This is a bug because the user needs to press it twice for the onPress callback to fire. But at least the user is not left with a bad visual state. Obviously we want a real fix for this, and when we have one, we can switch back to TouchableOpacity

### DIFF
--- a/packages/kit/src/client/DraggableFlatList.tsx
+++ b/packages/kit/src/client/DraggableFlatList.tsx
@@ -3,9 +3,9 @@ import DraggableFlatList, {
   DraggableFlatListProps,
   ScaleDecorator,
 } from 'react-native-draggable-flatlist'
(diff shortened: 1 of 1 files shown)


grabbou:
> Use TouchableWithoutFeedback to mitigate the issue where the pressed state is stuck

Ha! That was driving me crazy, great to see this fixed!

grabbou:
Ah, I guess not fully fixed, but at least it's not as confusing!

grabbou (approved):

PR #214, opened 2024-08-14
feat: kit: AnimatedProgress component

ericvicenti:
New component for the kit: AnimatedProgress which shows a progress bar that is animated on the client according to props from the server

This first implementation isn't great, ideally it would use reanimated or something for better performance

### DIFF
--- a/packages/kit/src/client/AnimatedProgress.tsx
+++ b/packages/kit/src/client/AnimatedProgress.tsx
@@ -1,0 +1,50 @@
+import { useEffect, useRef, useState } from 'react'
+import { Progress } from 'tamagui'
+import { z } from 'zod'
+
+export const animatedProgressPropsSchema = z.object({
+  startProgress: z.number().nullable(),
+  endProgress: z.number(),
+  duration: z.number(),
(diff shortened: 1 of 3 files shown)


grabbou:
> This first implementation isn't great, ideally it would use reanimated or something for better performance

I am wondering whether we should keep this outside of the kit and in the meantime, work towards adding support for higher-order components such as [Reanimated Layout Transitions](https://docs.swmansion.com/react-native-reanimated/docs/layout-animations/layout-transitions). 

I understand this is needed for the uprising project, but I am concern with increasing the surface area of the kit with components that are better to be outside of it. We should have some strategy as to what we put inside and what not, and maybe have a higher-level vision for it, as to what should be there (like a full list of components).

I actually think this could be a nice feature for some of our upcoming conferences!

grabbou:
After discussing this internally, we concluded that we're going to add this and future components to the kit based on the following strategy:
- are needed by a real-world application that is trying to deploy with playground
- might be useful for other apps
- cannot be implemented server-side
Please note some of them might be deprecated once we make Rise itself more powerful!

grabbou (approved):

PR #215, opened 2024-08-15
feat: allow rendering iterable elements on the client

grabbou:
Following up to previous PR, this adds ability to render Iterables, not just Arrays, as children and props. While our transportation mechanism currently does not support transporting anything but arrays, this makes it aligned with what server outputs, in case it gets passed directly as a prop.

Previously, this would throw an error that objects can't be rendered, which is correct, just not for iterables!

### DIFF
--- a/packages/react/src/__tests__/rise.test.tsx
+++ b/packages/react/src/__tests__/rise.test.tsx
@@ -69,6 +69,39 @@ it('should render an array of components', () => {
   `)
 })
 
+it('should render any iterable of components', () => {
+  const component = render(
+    <BaseRise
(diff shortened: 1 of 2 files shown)


ericvicenti:
Ok but the Set cannot be sent over the wire with the server and client. So this feels incomplete, and honestly I'm not sure anybody needs this feature 😅

grabbou:
It's incomplete, but it's a three-stage process.

Support on the server
Support on the client
One-day, support on the transportation layer

This is theoretically useful right now anyways, because you can skip the server and pass model state directly to Rise component (e.g. as a fallback) and that should align with what server can output and/or React accepts as a valid child.

PR #216, opened 2024-08-15
feat: open connection from QR Code without confirmation screen

grabbou:
https://github.com/user-attachments/assets/e4a886be-f992-4036-9cd5-aa8772535e93

There's little blank screen while transition happens. To make it go away, we need to listen to URLs with Linking API instead of using React Navigation and manually push the relevant screen.


grabbou:
blank screen will be likely fixed in a follow-up, but also, not a big deal for now
we will add error handling / preloading there too

ericvicenti (approved):
big step in the right direction!

we want the connect screen to handle loading indicator and error

PR #219, opened 2024-08-16
feat: set fallback when not connected to ws server

grabbou:


ericvicenti (approved):
great! we also want this for http data source

grabbou:
will try this locally in a different application and if it works, I guess we will be able to release a follow-up to npm.

PR #220, opened 2024-08-17
feat(jsx): add `jsx-dev-runtime`

grabbou:
When setting `jsxImportSource` to `@rise-tools/react`, certain bundlers may choose to use dev runtime (while in development). 

We must export `jsx-dev-runtime` with `jsxDev` and `jsxsDev`. We don't have any dev-specific behavior at this point, hence simple re-export.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -6,7 +6,8 @@
   "main": "./lib/index.js",
   "exports": {
     ".": "./lib/index.js",
-    "./jsx-runtime": "./lib/jsx-runtime.js"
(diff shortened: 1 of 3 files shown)


ericvicenti (approved):
`jsx-runtime.js` is supposed to be empty?

Otherwise, makes sense!

grabbou:
Oh LOL, definitely not. Should not be here at all! Initially, I wanted to support environments that didn't support package exports field (React Native right now), but I think it's minimal, I will skip it. 

We will document how to set it for React Native soon in a follow-up.

PR #221, opened 2024-08-17
fix: rendering Rise JSX from React Native / Metro context

grabbou:
Historically, we had this empty file so React Native can build (events had some references to node:crypto).

Now, it's not needed. We're actually rendering some Rise JSX inside React Native application (for fallback UI) and this is breaking it.


ericvicenti (approved):
Easy approve, I didn't even understand why it was here in the first place 😅

nice cleanup!

grabbou:
I had to look up old diff, 3 months ago things were super different!

PR #223, opened 2024-08-19
feat: add expo-video

grabbou:
This PR adds a new package that we will use for our conference demo.

The newly created <Video /> component contains props of VideoView from `expo-view` as well as certain props from `useVideoSource` hook / Player, that we set up for the user behind the scenes. That way, you use a single component to control that behavior. Will add documentation explaining this in a bit!

We currently do not support custom actions - this requires an improvement to actions() to access refs/context relative to the component that triggered action. We will likely have to look into this sooner than later.

Screenshots:

| Screen    | Screen | Screen |
| -------- | ------- | ------ |
|  Simulator Screenshot - iPhone 15 Pro - 2024-08-20 at 16 15 14 | Simulator Screenshot - iPhone 15 Pro - 2024-08-20 at 16 15 25 | Simulator Screenshot - iPhone 15 Pro - 2024-08-20 at 16 15 22


templates/full/src/controls/ui.tsx
```diff
@@ -129,6 +134,20 @@ function WebViewExample() {
   )
 }
 
+function VideoExample() {
+  return (
+    <>
+      <StackScreen title="Video" />
+      <Video
+        source="https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4"
```
grabbou, line 142: I took video from Expo documentation so no need to host it. We could consider uploading our own!

ericvicenti: yeah this is fine, its a VERY common example video

packages/kit-expo-video/src/video.tsx
```diff
@@ -0,0 +1,62 @@
...
+  preservesPitch,
+  showNowPlayingNotification,
+  staysActiveInBackground,
+  volume,
+  ...props
+}: { source: VideoSource; autoplay?: boolean } & VideoPlayerSettings &
+  Omit<VideoViewProps, 'player'>) {
+  const player = useVideoPlayer(source)
+
+  useEffect(() => {
```
ericvicenti, line 33: an effect? boo!

just kidding, not sure how else we would do this 😂

ericvicenti (approved):
ok looks good!

ericvicenti:
One thing is that the video view in the example component doesn't have the same aspect ratio as the video itself. Maybe thats an easy fix with something like `aspectRatio={9/16}`

grabbou:
Yeah, aspectRatio: 16/9 fixed it. Very elegant, thanks!

PR #224, opened 2024-08-21
feat: add `kit-react-native`

grabbou:
This package is not feature complete, just shipping components as that is needed for our conference talk demo.

- I moved FlatList to a dedicated package
- Added SectionList and VirtualizedList
- Removed special handling of "header" and "footer" props. Turns out "ListHeaderComponent" takes an element as well (which is what Rise passes down to children on the client). For some reason, that was not working before, or we have fixed and underlying issue in the meantime.
- Moved FlatList example under React Native example (new entry in UI controls)

This is a breaking change, but since we're `0` major, I don't think we should bump it as such. 

Breaking change is:
- FlatList is now in kit-react-native
- header is now ListHeaderComponent (as per React Native)


grabbou:
@ericvicenti should we showcase all elements? or none? 

Simulator Screenshot - iPhone 15 Pro - 2024-08-21 at 10 56 45

grabbou:
Question:
- Should we keep all relevant examples (FlatList, ActivityIndicator) top-level, alongside e.g. DraggableFlatList or BottomSheet in the UI controls, or... group them under "Base React Native Elements"?
- Should we document all base React Native elements? They're kinda obvious on one hand!

packages/kit-react-native/src/lists.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import { FlatList, FlatListProps, SectionList, SectionListProps } from 'react-native'
+
+type Item = { key: string; label: React.ReactElement }
+
+export function RNFlatList(props: Omit<FlatListProps<Item>, 'keyExtractor' | 'renderItem'>) {
+  return (
+    <FlatList {...props} keyExtractor={(item) => item.key} renderItem={({ item }) => item.label} />
```
grabbou, line 8: As a follow-up, we should do the same for DraggableFlatList before cutting next release - remove `header` in favour of just `ListHeaderComponent`.

Turns out that the following:
```
ListHeaderComponent={props.header ? () => props.header : undefined}
```
is not necessary, as we can just pass an element directly.

I believe we may have had another kind of an issue before inside Rise renderer which prompted this workaround, but since that issue is now gone, this can also be simplified.

ericvicenti: I'm not sure `item.label` is the best name, because this view is more than a label. Just `item` makes sense but then the implementation here would be `item.item` which seems confusing. Maybe `item.view`?

templates/full/src/controls/ui.tsx
```diff
@@ -125,6 +125,53 @@ function QRCodeExample() {
...
+        <Button theme="blue" marginVertical="$2">
+          Rise Tools
+        </Button>
+      ),
+    },
+  ]
+  return (
+    <>
+      <StackScreen title="QR Code" />
+      <YStack gap="$8" padding="$4">
```
ericvicenti, line 158: if we are putting all RN examples here in one place, we are missing a lot of things. I think at least the lists should be separate so we can have a list view here and show more of the components

templates/full/src/controls/ui.tsx
```diff
@@ -125,6 +125,53 @@ function QRCodeExample() {
...
+      label: (
+        <Button theme="blue" marginVertical="$2">
+          Rise Tools
+        </Button>
+      ),
+    },
+  ]
+  return (
+    <>
+      <StackScreen title="QR Code" />
```
ericvicenti, line 157: wrong title

templates/full/src/controls/ui.tsx
```diff
@@ -53,6 +52,7 @@ export const models = {
   qrcode: QRCodeExample,
   webview: WebViewExample,
   video: VideoExample,
+  rn: ReactNativeExample,
```
ericvicenti, line 55: maybe we should have an example which shows the combination of `StackScreen` and `StatusBar` features? Although that example might need to be dynamic somehow either with server or client state.. just an idea, not important for this PR

ericvicenti (approved):
Good stuff!

One question, is `kit-react-native` part of "the kit"? If so, what do you get when you import `View`? RN View or Tamagui View?

packages/kit-react-native/src/lists.tsx
grabbou: That's fine, but would be breaking to previous FlatList users and DraggableList users if we change it! But I am down, we should simplify it

grabbou:
> One question, is kit-react-native part of "the kit"? If so, what do you get when you import View? RN View or Tamagui View?

Good question. Right now, I prefixed it with RN, like we did with SVG, b/c View was clashing with Tamagui. I think React Native should either not be part of the kitchen sink, or be exported behind a namespace, as we're Tamagui-first.

Thoughts?

templates/full/src/controls/ui.tsx
grabbou: Ok, I'll revert List back to be top-level, I think rest we can showcase later when we have better idea how to do it! (kinda something else on my mind now)

PR #226, opened 2024-08-21
refactor: remove `header` and `footer` props from DraggableFlatList

grabbou:
This aligns DraggableFlatList with FlatList props, no special treatment for header/footer is required. This must've been a workaround back when we had some issues within renderer itself.


ericvicenti (commented):
This will break anybody who is currently using the old props. Plus I prefer the simple names

grabbou:
Closing this. Will re-open with deprecations a new PR

PR #227, opened 2024-08-21
fix: warning about missing keys when using server-side helpers

grabbou:
Aside from fixing the warning itself, I tried to simplify and make the naming inside JSX less confusing. This file is evolving (as JSX support is still experimental) and I felt like certain things did not make much sense anymore!

Mainly:
- Fixed wrong comment (a bit misleading) as far as what is the return type of `componentFactory`
- Removed `RiseElement` and `ReactNode` types in favour of `ServerModelState`. That allowed to remove type casting and other workarounds that were present in this PR in earlier commits


ericvicenti (approved):
looks great

PR #229, opened 2024-08-22
fix: proper handling of static children in dev runtime

grabbou:
jsxDEV has different signature, there's no `jsxsDEV`, it takes `isStatic` as an additional boolean argument. We can't just export two factories, one for dynamic, second for static, we have to handle it differently, hence the change.

FWIW the handling is very similar to how React does this internally now! 

https://github.com/facebook/react/blob/main/packages/react/src/jsx/ReactJSXElement.js#L490-L501

(note explicit boolean)

```tsx
const profile = (
  <div>
    <img src="avatar.png" className="profile" />
    <h3>{[user.firstName, user.lastName].join(" ")}</h3>
  </div>
);
```

```tsx
import { jsxDEV as _jsxDEV } from "react/jsx-dev-runtime";

const _jsxFileName = "input.jsx";
const profile = _jsxDEV("div", {
  children: [
    _jsxDEV("img", {
      src: "avatar.png",
      className: "profile",
    }, undefined, false, { fileName: _jsxFileName, lineNumber: 3, columnNumber: 5 }, this),
    _jsxDEV("h3", {
      children: [user.firstName, user.lastName].join(" "),
    }, undefined, false, { fileName: _jsxFileName, lineNumber: 4, columnNumber: 5 }, this),
  ]},
  undefined, false, { fileName: _jsxFileName, lineNumber: 2, columnNumber: 3 }, this
);
```

### DIFF
--- a/packages/react/src/jsx-runtime.tsx
+++ b/packages/react/src/jsx-runtime.tsx
@@ -13,13 +13,6 @@ import {
 
 type JSXFactory = (
   /**
-   * When rendering fragments, `componentFactory` will be undefined
-   * ```tsx
-   * <>
-   *   <View>foo</View>
-   * </>
-   * ```
-   *
    * When rendering server-side component definitions, `componentFactory` will return `ReactElement`:
    * ```tsx
    * const View = createComponentDefinition('View')
@@ -39,7 +32,7 @@ type JSXFactory = (
    * <Helper />
    * ```
    */
-  componentFactory: ((props: any) => ServerModelState | ReactElement) | undefined,
+  componentFactory: (props: any) => ServerModelState | ReactElement,
   { children, ...passedProps }: Record<string, any>,
   key?: string
 ) => ServerModelState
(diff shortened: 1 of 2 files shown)


packages/react/src/jsx-runtime.tsx
```diff
@@ -34,18 +34,21 @@ type JSXFactory = (
...
-  Object.defineProperty(passedProps.children, '$static', {
-    value: true,
-    enumerable: false,
-  })
-  return jsx(componentFactory, passedProps, key)
+  return jsx(componentFactory, passedProps, key, true)
 }
 
-export const jsx: JSXFactory = (componentFactory, passedProps, key) => {
+export const jsx: JSXFactory = (componentFactory, passedProps, key, isStaticChildren) => {
```
grabbou, line 45: This PR is ready for review, but I need to change this to `jsxImpl`, as `jsx` itself as called by Babel or other transformers does not receive this boolean argument, so this is a bit misleading.

We will have jsx, jsxs, jsxDEV and jsxImpl (the base function), just like React.

ericvicenti (approved):
I only understand this 50% but I'm a glass-half-full kind of guy

PR #230, opened 2024-08-24
fix: state mutation setter allows silent updates

ericvicenti:
One of the main purposes of the state setter function is to allow the mutation to return null without emitting an update to subscribers

Previously the subscribers would always be called, this makes it so you can silently call setState with the same value

### DIFF
--- a/packages/server/src/__tests__/model-state.test.ts
+++ b/packages/server/src/__tests__/model-state.test.ts
@@ -15,6 +15,26 @@ describe('model-state', () => {
     setS((s) => s + 1)
     expect(s.get()).toBe(1)
   })
+  test('mutation setter with no update', () => {
+    const handler = vi.fn()
+    const [s, setS] = state(0)
+    const release = s.subscribe(handler)
+    expect(handler).toBeCalledTimes(1)
+    expect(handler).toHaveBeenLastCalledWith(0)
+    expect(s.get()).toBe(0)
+    setS((s) => s)
+    expect(handler).toBeCalledTimes(1)
+    setS((s) => s + 1)
+    expect(handler).toBeCalledTimes(2)
+    expect(handler).toHaveBeenLastCalledWith(1)
+    release()
+    setS((s) => s + 1)
+    expect(s.get()).toBe(2)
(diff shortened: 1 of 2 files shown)


packages/server/src/__tests__/model-state.test.ts
```diff
@@ -15,6 +15,26 @@ describe('model-state', () => {
...
+    expect(handler).toHaveBeenLastCalledWith(0)
+    expect(s.get()).toBe(0)
+    setS((s) => s)
+    expect(handler).toBeCalledTimes(1)
+    setS((s) => s + 1)
+    expect(handler).toBeCalledTimes(2)
+    expect(handler).toHaveBeenLastCalledWith(1)
+    release()
+    setS((s) => s + 1)
+    expect(s.get()).toBe(2)
```
grabbou, line 32: What happens in between setS invocations that s is always 2? Does "release" function resets it in some way?

grabbou (approved):
Sounds good. Nice optimization!

packages/server/src/__tests__/model-state.test.ts
```diff
@@ -15,6 +15,26 @@ describe('model-state', () => {
...
   })
+  test('mutation setter with no update', () => {
+    const handler = vi.fn()
+    const [s, setS] = state(0)
+    const release = s.subscribe(handler)
+    expect(handler).toBeCalledTimes(1)
+    expect(handler).toHaveBeenLastCalledWith(0)
+    expect(s.get()).toBe(0)
+    setS((s) => s)
+    expect(handler).toBeCalledTimes(1)
```
ericvicenti, line 26: I messed up the comments in this test. THIS is the main thing that we are checking, that the handler does not get called again after doing `setS(s => s)`

packages/server/src/__tests__/model-state.test.ts
ericvicenti: s starts off as 0 and we increment it twice, that is why s is 2 here

release in this context means "unsubscribe" so we also make sure that the subscription doesn't get called anymore. but that isn't really necessary in this test

PR #232, opened 2024-08-26
feat: deprecate `header` and `footer` on DraggableFlatList

grabbou:
Things brings it to feature parity with FlatList and other lists as well as DraggableFlatList itself. Deprecations will be removed in the upcoming releases, so please migrate if you're relying on this!

### DIFF
--- a/packages/kit/src/client/DraggableFlatList.tsx
+++ b/packages/kit/src/client/DraggableFlatList.tsx
@@ -8,10 +8,7 @@ import { TouchableWithoutFeedback } from 'react-native-gesture-handler'
 type Item = { key: string; label: React.ReactElement; onPress?: () => void }
 
 export function RNDraggableFlatList(
-  props: Omit<
-    DraggableFlatListProps<Item>,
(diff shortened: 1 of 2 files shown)


ericvicenti (commented):
kind of confusing that the prop is called this if elements are accepted. the convention is lowercase and not "Component"

so I actually prefer `header` and `footer`

grabbou:
Yeah, well, it's how the prop is called on FlatList (that DraggableFlatList uses under the hood). I thought we didn't want to deviate from the underlying component, hence the PR.

ericvicenti:
idk, the API for DraggableFlatList is already different because the is no `renderItem`, so its no big deal I think to deviate in places where we can make improvements.

grabbou:
Yeah, we don't have `renderItem` and `keyExtractor`. I get your point, but these props were dropped out of necessity. I would stick to original props as much as possible, but if you have strong feelings about it, I am fine leaving it as is. Let me know 🙏

PR #233, opened 2024-08-26
refactor: remove `localState` experimental until it's ready

grabbou:
Removing this, we will reimplement with React 19. Right now, its implementation is slowing perf. of Rise and not providing much benefit. 

In the future, we will look back into this PR as a reference on how things should be implemented.


grabbou:
This is currently a draft. I am looking at making our view tree lean, so that properties such as `asChild` work out of the box.

